PR #9, opened 2020-01-09
Update deps

### LINKED ISSUE #10: Update to CSI external-snapshotter 2.0.0
To get on track with the latest release, we need to be sure we update our dependency from the release candidate from 2.0.0-rc4 to 2.0.0, and ensure that the Kubernetes dependencies are also updated to match.

nrb:
Update to CSI external-snapshotter 2.0.0 and a compatible version of kubernetes deps

Closes #10


Gopkg.lock
```diff
@@ -210,6 +207,20 @@
   revision = "298182f68c66c05229eb03ac171abe6e309ee79a"
   version = "v1.0.3"
 
+[[projects]]
+  digest = "1:4c49f4b3261fcd083db9c566038401a642e7d59f95418be204f43cb47cb3fd61"
+  name = "github.com/vmware-tanzu/velero"
```
skriss: it looks like we're getting a duplicate entry for velero -- one for `github.com/heptio/velero` and one for `github.com/vmware-tanzu/velero`. Guessing that updating `Gopkg.toml` to use the `vmware-tanzu` ref, and ensuring any code refs use that as an import path, will resolve this

nrb: Good catch, I'll do a find/replace on that.

ashish-amarnath: was this repo not being built or were the `github.com/heptio/velero/...` being taken care of by github redirects?

skriss: redirects are in place so i'm not sure that it would've blocked anything from building, it was just a cleanup thing

Issue #1, opened 2019-07-24
PVC restore fails due to "Snapshot is not Ready"


betta1:
I installed the velero-csi-plugin and followed the steps to backup and restore the demo app. I have the ebs-csi-driver up and running on my k8s on AWS. PV snapshots via the velero csi plugin work as expected, however PVC restores from the csi snapshots fail since volume snapshots are not ready. This causes the pod to be stuck in pending state:

"Warning  ProvisioningFailed    18s (x6 over 47s)  ebs.csi.aws.com_ebs-csi-controller-0_178686e9-ae29-11e9-af00-6e03acfc45b2  failed to provision volume with StorageClass "ebs-sc": error getting snapshot handle for snapshot velero-ebs-claim-qrtd4: snapshot velero-ebs-claim-qrtd4 is not Ready"

$ velero restore create --from-backup demo-backup
    Restore request "demo-backup-20190724104629" submitted successfully.
    ...
$ kubectl get po -n demo
   NAME   READY   STATUS    RESTARTS   AGE
   app        0/1        Pending   0                 12m

$ kubectl describe pvc ebs-claim -n demo
Name:          ebs-claim
Namespace:     demo
StorageClass:  ebs-sc
Status:        Pending
.
.
.
Type     Reason                Age                From                                                                       Message
----     ------                ----               ----                                                                       -------
Normal   Provisioning          18s (x6 over 47s)  ebs.csi.aws.com_ebs-csi-controller-0_178686e9-ae29-11e9-af00-6e03acfc45b2  External provisioner is provisioning volume for claim "demo/ebs-claim"
Warning  ProvisioningFailed    18s (x6 over 47s)  ebs.csi.aws.com_ebs-csi-controller-0_178686e9-ae29-11e9-af00-6e03acfc45b2  failed to provision volume with StorageClass "ebs-sc": error getting snapshot handle for snapshot velero-ebs-claim-qrtd4: snapshot velero-ebs-claim-qrtd4 is not Ready
Normal   ExternalProvisioning  9s (x5 over 47s)   persistentvolume-controller                                                waiting for a volume to be created, either by external provisioner "ebs.csi.aws.com" or manually created by system administrator

nrb:
Thanks for the report @betta1. I will continue to test this with the new beta types and see if we can catch this.

skriss:
xref https://github.com/vmware-tanzu/velero/issues/1799 (similar issue with legacy VolumeSnapshotters)

[...]

PR #16, opened 2020-02-19
Add CSI volumesnapshot capability into Velero

### LINKED ISSUE #5: Don't back up a PVC already backed up by restic
From the [design document](https://github.com/vmware-tanzu/velero/blob/master/design/csi-snapshots.md#restic-and-csi-interaction):

Volumes found in a Pod's backup.velero.io/backup-volumes list will use Velero's current Restic code path. This also means Velero will continue to offer Restic as an option for CSI volumes.

The velero.io/csi-pvc BackupItemAction plugin will inspect pods in the namespace to ensure that it does not act on PVCs already being backed up by restic.

This is preferred to modifying the PVC due to the fact that Velero's current backup process backs up PVCs and PVs mounted to pods at the same time as the pod.

A drawback to this approach is that we're querying all pods in the namespace per PVC, which could be a large number. In the future, the plugin interface could be improved to have some sort of context argument, so that additional data such as our existing resticSnapshotTracker could be passed to plugins and reduce work.

ashish-amarnath:

fixes https://github.com/vmware-tanzu/velero/issues/1566
fixes #5
fixes https://github.com/vmware-tanzu/velero/issues/2139


velero-csi-plugin/csi_snapshotter.go
```diff
@@ -23,9 +23,11 @@ import (
 	"github.com/sirupsen/logrus"
 
 	corev1api "k8s.io/api/core/v1"
+
```
skriss: nit: rm blank line

velero-csi-plugin/csi_snapshotter.go
```diff
@@ -34,6 +36,11 @@ import (
 	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
 )
 
+const (
+	veleroBackupNameAnnotation   = "velero.io/backup-name"
```
skriss: this constant is defined in the velero `pkg/apis/velero/v1` package, can probably reference it from there

velero-csi-plugin/csi_snapshotter.go
```diff
@@ -34,6 +36,11 @@ import (
 	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
```
skriss: nit: let's alias the `pkg/apis/velero/v1` package as `velerov1api` per our standard.

velero-csi-plugin/csi_snapshotter.go
```diff
@@ -55,42 +62,58 @@ func (p *CSISnapshotter) AppliesTo() (velero.ResourceSelector, error) {
 func (p *CSISnapshotter) Execute(item runtime.Unstructured, backup *v1.Backup) (runtime.Unstructured, []velero.ResourceIdentifier, error) {
 	p.log.Info("CSISnapshotter Execute")
 
+	// Do nothing if volume snapshots have not been requested in this backup
+	if !*backup.Spec.SnapshotVolumes {
```
skriss: need to nil-check this. we have a `pkg/util/boolptr` in velero, if you want to use it

velero-csi-plugin/csi_snapshotter.go
```diff
@@ -55,42 +62,58 @@ func (p *CSISnapshotter) AppliesTo() (velero.ResourceSelector, error) {
...
+		p.log.Infof("Volume snpashotting not requested for backup %s/%s", backup.Namespace, backup.Name)
```
skriss:
```suggestion
		p.log.Infof("Volume snapshotting not requested for backup %s/%s", backup.Namespace, backup.Name)
```

velero-csi-plugin/util.go
```diff
@@ -0,0 +1,143 @@
...
+func contains(slice []string, key string) bool {
+	for _, i := range slice {
+		if i == key {
+			return true
+		}
+	}
+	return false
+}
+
+func isPVCBackedupByRestic(pvcNamespace, pvcName string, podClient corev1client.PodsGetter) (bool, error) {
```
skriss:
```suggestion
func isPVCBackedUpByRestic(pvcNamespace, pvcName string, podClient corev1client.PodsGetter) (bool, error) {
```

velero-csi-plugin/csi_snapshotter.go
```diff
@@ -117,10 +140,40 @@ func (p *CSISnapshotter) Execute(item runtime.Unstructured, backup *v1.Backup) (
...
 		annotations = make(map[string]string)
 	}
-	annotations["velero.io/volume-snapshot-name"] = upd.Name
+	annotations[veleroSnapshotNameAnnotation] = upd.Name
+	annotations[veleroBackupNameAnnotation] = backup.Name
 	accessor.SetAnnotations(annotations)
 
-	return item, nil, nil
+	//TODO: add retry logic here. CSI driver may take time to reconcile the snapshotcontents object
+	snapshotContents, err := getVolumeSnapshotContentForVolumeSnapshot(snapshotClient.SnapshotV1beta1(), upd.Name)
```
nrb: Does this mean that we're pausing until the contents are completely ready?

I don't think we need to do that, do we? We simply need the pointer to the contents object, so we know what to fetch and include.

velero-csi-plugin/csi_snapshotter.go
```diff
@@ -117,10 +140,40 @@ func (p *CSISnapshotter) Execute(item runtime.Unstructured, backup *v1.Backup) (
...
+
+	// TODO: Create GroupResource obejects for volumesnapshot objects
+	additionalItems := []velero.ResourceIdentifier{
+		{
+			GroupResource: schema.GroupResource{
```
nrb: We could probably define these in core Velero's `pkg/kuberesource` package.

velero-csi-plugin/csi_snapshotter.go
nrb: The standard is documented at https://velero.io/docs/master/code-standards/#imports

velero-csi-plugin/csi_snapshotter.go
ashish-amarnath: Yes, that was the idea. But, I put it here for now.

velero-csi-plugin/csi_snapshotter.go
ashish-amarnath: The retry logic here is for the CSI driver to create the snapshotcontent object. This querying of the snapshotcontent object can occur before the CSI driver has had a change to create the snapshotcontent object.

Also, I think we should wait for `VolumeSnapshot.Status.readyToUse` to be true. If we don't, then it is possible that the velero backup process, while backing up the volumesnapshot as the additional items, can make progress without the CSI driver having completed its snapshotting. Leading to the possibility that we backup a Volumesnapshot that is not ready to be use, i.e., not ready to be restored. Which is what may be causing https://github.com/vmware-tanzu/velero-plugin-for-csi/issues/1

ashish-amarnath:
@skriss This PR is just for the backup process. I agree that we don't want one giant PR for all the CSI work 🙂

velero-csi-plugin/csi_snapshotter.go
nrb: > Also, I think we should wait for VolumeSnapshot.Status.readyToUse to be true.

We can discuss this in a separate issue, as I don't want to lump all work into this PR.

However, I will say that the design of CSI is such that the intention is _not_ to pause all backup work until `readyToUse == true`. This is also similar to how Velero currently works, by launching an async request against the storage provider's snapshot APIs, then proceeding.

While I'm aware that this can get into the state that it's not ready to restore, the restoration state should be reconciled when the `VolumeSnapshot` and `VolumeSnapshotContent` objects are recreated in the new cluster. Their `Status` fields will be cleared upon restoration by Velero, and the on-disk snapshot itself will be queried for readiness. This would be considered [an import case](https://kubernetes.io/blog/2018/10/09/introducing-volume-snapshot-alpha-for-kubernetes/#importing-an-existing-snapshot-with-kubernetes) in the upstream terminology.

nrb:
>  This PR is to address vmware-tanzu/velero#1566 which is just the backup bit.

This appears to also be covering #5.

velero-csi-plugin/csi_snapshotter.go
ashish-amarnath: Yeah, we can discuss whether or not we should wait for `VolumeSnapshot.Status.readyToUse` in a separate issue.

skriss:
As well as #2139 :)

Not a big deal - we'll be fine on this PR - just wanted to make sure we're on the same page going forward.

velero-csi-plugin/csi_snapshotter.go
skriss: This is an existing issue with Velero & non-CSI snapshots too -- see https://github.com/vmware-tanzu/velero/issues/1799 - so I'd like to think about a solution for both.

velero-csi-plugin/csi_snapshotter.go
```diff
@@ -45,55 +54,75 @@ type CSISnapshotter struct {
 func (p *CSISnapshotter) AppliesTo() (velero.ResourceSelector, error) {
 	p.log.Info("CSISnapshotter AppliesTo")
 
+	//TODO: Update this to look at only PVCs with "magic" label
```
skriss: can you clarify what this means? I'm not clear on what the "magic" label is

velero-csi-plugin/csi_snapshotter.go
```diff
@@ -45,55 +54,75 @@ type CSISnapshotter struct {
...
-		return nil, nil, errors.Wrap(err, "error getting storage class")
+		return nil, nil, errors.WithStack(err)
 	}
-
-	snapshotClasses, err := snapshotClient.SnapshotV1beta1().VolumeSnapshotClasses().List(metav1.ListOptions{})
+	if isResticUsed {
+		p.log.Infof("Nothing to do for PVC %s/%s is backed up using restic")
+		return item, nil, nil
+	}
+	p.log.Debugf("Fetching underlying PV of PVC %s", fmt.Sprintf("%s/%s", pvc.Namespace, pvc.Name))
```
skriss: nit: add blank line between L92 and L93

velero-csi-plugin/csi_snapshotter.go
```diff
@@ -45,55 +54,75 @@ type CSISnapshotter struct {
...
+	// Do nothing if restic is used to backup this PV
+	isResticUsed, err := isPVCBackedUpByRestic(pvc.Namespace, pvc.Name, client.CoreV1())
 	if err != nil {
-		return nil, nil, errors.Wrap(err, "error getting storage class")
+		return nil, nil, errors.WithStack(err)
 	}
-
-	snapshotClasses, err := snapshotClient.SnapshotV1beta1().VolumeSnapshotClasses().List(metav1.ListOptions{})
+	if isResticUsed {
+		p.log.Infof("Nothing to do for PVC %s/%s is backed up using restic")
```
skriss: Nit: let's say something like "Skipping CSI snapshot of PVC %s/%s because it's being backed up with restic" to be a little more clear.

Although, now that I'm reading L95-102, I guess we don't even know that this *is* a CSI PVC yet, so my suggestion may also be misleading.

Perhaps we should do the CSI check first, so we can exit right away if it's not a CSI volume. WDYT?

velero-csi-plugin/csi_snapshotter.go
```diff
@@ -45,55 +54,75 @@ type CSISnapshotter struct {
...
 	}
 
-	var snapshotClass *snapshotv1beta1api.VolumeSnapshotClass
-	for i := range snapshotClasses.Items {
-		if snapshotClasses.Items[i].Driver == storageClass.Provisioner {
-			snapshotClass = &snapshotClasses.Items[i]
-		}
+	// no storage class: we don't know how to map to a VolumeSnapshotClass
+	if pvc.Spec.StorageClassName == nil {
+		return item, nil, errors.Errorf("Cannot snpashot PVC %s/%s, PVC has no storage class.", pvc.Namespace, pvc.Name)
```
skriss: s/snpashot/snapshot

velero-csi-plugin/csi_snapshotter.go
```diff
@@ -117,10 +147,61 @@ func (p *CSISnapshotter) Execute(item runtime.Unstructured, backup *v1.Backup) (
...
 		annotations = make(map[string]string)
 	}
-	annotations["velero.io/volume-snapshot-name"] = upd.Name
+	annotations[veleroSnapshotNameAnnotation] = upd.Name
+	annotations[velerov1api.BackupNameLabel] = backup.Name
 	accessor.SetAnnotations(annotations)
 
-	return item, nil, nil
+	var snapshotContent *snapshotv1beta1api.VolumeSnapshotContent
+	p.log.Info("Fetching volumesnapshotconent for volumesnapshot")
```
skriss: s/volumesnapshotconent/volumesnapshotcontent

velero-csi-plugin/util.go
```diff
@@ -0,0 +1,143 @@
+package main
```
skriss: need to add a copyright header (year = 2020)

velero-csi-plugin/util.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		for _, p := range podList.Items {
+			for _, v := range p.Spec.Volumes {
+				if v.PersistentVolumeClaim != nil && v.PersistentVolumeClaim.ClaimName == pvcName {
+					podsUsingPVC = append(podsUsingPVC, p)
+				}
+			}
+		}
+		if podList.Continue == "" {
```
skriss: this might be a little confusing given that we're not actually using a limit on the List call - let's simplify to just do a single list and iterate through it, rather than having the outer loop.

velero-csi-plugin/util.go
```diff
@@ -0,0 +1,143 @@
...
+		snapshotClasses, err := snapshotGetter.VolumeSnapshotClasses().List(metav1.ListOptions{})
+		if err != nil {
+			return nil, errors.Wrap(err, "error listing volumesnapshot classes")
+		}
+		for _, sc := range snapshotClasses.Items {
+			if sc.Driver == provisioner {
+				return &sc, nil
+			}
+		}
+		if snapshotClasses.Continue == "" {
```
skriss: this might be a little confusing given that we're not actually using a limit on the List call - let's simplify to just do a single list and iterate through it, rather than having the outer loop.

velero-csi-plugin/util.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func getVolumeSnapshotClassForStorageClass(provisioner string, snapshotGetter snapshotter.VolumeSnapshotClassesGetter) (*snapshotv1beta1api.VolumeSnapshotClass, error) {
+	for {
+		snapshotClasses, err := snapshotGetter.VolumeSnapshotClasses().List(metav1.ListOptions{})
+		if err != nil {
+			return nil, errors.Wrap(err, "error listing volumesnapshot classes")
+		}
+		for _, sc := range snapshotClasses.Items {
+			if sc.Driver == provisioner {
```
skriss: I *believe* that the provisioner to snapshotclass is 1:many, meaning we could have more than one snapshot class for a given provisioner. We'll need to handle this somehow. Is there a concept of a default class per provisioner? If so, then we probably want to do something like:

- if there's only one snapshot class for the provisioner, use it
- if there's more than one snapshot class for the provisioner, use the default (if this is a thing)
- support some kind of annotation/label to allow the user to specify a non-default snapshot class to be used for a given PVC

We can refine this in a follow-up PR - let's just create a separate issue for it

velero-csi-plugin/util.go
```diff
@@ -0,0 +1,143 @@
...
+			return nil, errors.WithStack(err)
+		}
+
+		for _, sc := range snapshotContents.Items {
+			if sc.Spec.VolumeSnapshotRef.APIVersion == snapshotv1beta1api.SchemeGroupVersion.String() &&
+				sc.Spec.VolumeSnapshotRef.Name == volSnapName {
+				return &sc, nil
+			}
+		}
+		if snapshotContents.Continue == "" {
```
skriss: this might be a little confusing given that we're not actually using a limit on the List call - let's simplify to just do a single list and iterate through it, rather than having the outer loop.

velero-csi-plugin/util.go
```diff
@@ -0,0 +1,143 @@
...
+		if snapshotClasses.Continue == "" {
+			break
+		}
+	}
+	return nil, errors.Errorf("failed to get volumesnapshotclass for provisioner %s", provisioner)
+}
+
+func getVolumeSnapshotContentForVolumeSnapshot(volSnapName string, snapshotContentsGetter snapshotter.VolumeSnapshotContentsGetter) (*snapshotv1beta1api.VolumeSnapshotContent, error) {
+	for {
+		snapshotContents, err := snapshotContentsGetter.VolumeSnapshotContents().List(metav1.ListOptions{})
```
nrb: When would there be more than 1 `VolumeSnapshotContent` for a given `VolumeSnapshot`?

velero-csi-plugin/csi_snapshotter.go
nrb: If the "magic" label has to do with restic selection, see this discussion on the CSI design proposal - it's tricky to get the ordering right :( https://github.com/vmware-tanzu/velero/pull/1661#discussion_r315375468

velero-csi-plugin/csi_snapshotter.go
```diff
@@ -17,6 +17,9 @@ limitations under the License.
 package main
```
nrb: We'll need to get the copyright statement added here, too.

velero-csi-plugin/csi_snapshotter.go
ashish-amarnath: I wasn't able to quickly come up w/ a way to describe that label. Hence the "magic". But the idea was to not have this plugin run for all PVCs, but just on the ones that have this label. I was not convinced we needed this label and hence the vague comment. 
I hadn't read that comment before. Based on what I read there, and the fact that there are checks we run to only operate on CSI PVs. I am convinced that the label is not required and will remove this comment also.

velero-csi-plugin/csi_snapshotter.go
ashish-amarnath: I agree w/ changing the order of the checks. The `isCSIPV` check is less expensive of the two calls.

velero-csi-plugin/csi_snapshotter.go
nrb: Yeah, I agree that if we can not have this plugin run all the time that would be ideal. I think moving the CSI check up and exiting the plugin early if it's not a CSI volume is probably the best we can do for now. That way, we at least minimize the work the plugin does.

velero-csi-plugin/util.go
ashish-amarnath: I had understood the `List` API incorrectly.
Yeah removing the outer loop makes sense,

velero-csi-plugin/csi_snapshotter.go
ashish-amarnath: marking this as resolved.

velero-csi-plugin/util.go
nrb: I'm asking because usually the relationship is, as I understand, usually a 1:1 relationship, much like a PVC to a PV.

skriss: isn't there a two-way binding here, such that we could just repeatedly `Get` the VolumeSnapshot and wait for the relevant binding field/ref on the `VolumeSnapshot` to be set, rather than having to list all VolumeSnapshotContents?

ashish-amarnath: Just confirmed from the backup yaml, the links are all 1-way.
`volumesnapshotcontents.snapshot`  has the name of `volumeSnapshotClass` and object ref to `VolumeSnapshot`
`volumesnapshots` just has the name of the `volumeSnapshotClass`

ashish-amarnath: I take that back.
The `volumesnapshots.snapshot.storage.k8s.io`  has a `boundVolumeSnapshotContentName` field with the `volumesnapshotcontent` name

nrb: Yeah, the binding should definitely be 2 way.

Via [`VolumeSnapshot.Status.BoundSnapshotContentName`](https://github.com/kubernetes-csi/external-snapshotter/blob/master/pkg/apis/volumesnapshot/v1beta1/types.go#L115) on the backup side and [`VolumeSnapshot.Spec.Source.VolumeSnapshotContentName`](https://github.com/kubernetes-csi/external-snapshotter/blob/master/pkg/apis/volumesnapshot/v1beta1/types.go#L102) on the restore side.

The difference between the two will be that `BoundVolumeSnapshotContentName` will be created in response to a request for the CSI controllers to make a snapshot. `VolumeSnapshotContentName` will be used in the "import" case when a storage-side snapshot already exists and the user or some program is constructing a Kubernetes representation of it.

ashish-amarnath: We can simplify this code to get the `volumesnapshot` that we just created and read the `status.BOundVolumeSnapshotContentName`

velero-csi-plugin/util.go
ashish-amarnath: Good catch! Yes, it is possible to have multiple snapshotclasses for the same provisioner, say with different `VolumesnapshotClass.Parameters` or `VolumesnapshotClass.DeletionPolicy`. I'll add a comment that we'll address in another PR.
Created https://github.com/vmware-tanzu/velero-plugin-for-csi/issues/17 for that

velero-csi-plugin/csi_snapshotter.go
ashish-amarnath: Opened this PR https://github.com/vmware-tanzu/velero/pull/2288

skriss:
@nrb this one's just waiting for an LGTM from you.

velero-csi-plugin/util.go
```diff
@@ -0,0 +1,141 @@
...
+	for {
+		vs, err := snapshotClient.VolumeSnapshots(volSnap.Namespace).Get(volSnap.Name, metav1.GetOptions{})
+		if err != nil {
+			return nil, errors.Wrapf(err, fmt.Sprintf("failed to get volumesnapshot %s/%s", volSnap.Namespace, volSnap.Name))
+		}
+
+		// TODO: add timeout
+		if vs.Status == nil || vs.Status.BoundVolumeSnapshotContentName == nil {
+			log.Infof("Waiting for CSI driver to reconcile volumesnapshot %s/%s. Retrying in 5s", volSnap.Namespace, volSnap.Name)
+			time.Sleep(5 * time.Second)
```
nrb: I don't think we should hold up this PR, but instead of sleeps, this could probably use the `PollImmediate` function, like at https://github.com/vmware-tanzu/velero/blob/master/pkg/restore/restore.go#L636

velero-csi-plugin/util_test.go
```diff
@@ -0,0 +1,851 @@
...
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+var (
+	csiStorageClass = "csi-hostpath-sc"
+	volumeMode      = "Filesystem"
+)
+
+func TestGetPVForPVC(t *testing.T) {
+	boundPVC := &corev1api.PersistentVolumeClaim{
```
nrb: Not going to hold up this PR for it, especially since I don't think any of us have mentioned it and because it's in a separate repo but we have a library of "builders" in the Velero codebase that we've been trying to standardize on.

They're at https://github.com/vmware-tanzu/velero/tree/master/pkg/builder

PR #18, opened 2020-02-29
Implement VolumeSnapshot restore

ashish-amarnath:

Fixes: https://github.com/vmware-tanzu/velero/issues/2141 https://github.com/vmware-tanzu/velero/issues/2140


skriss:
👀

velero-csi-plugin/csi_restorer.go
```diff
@@ -63,16 +66,35 @@ func (p *CSIRestorer) Execute(input *velero.RestoreItemActionExecuteInput) (*vel
 		return nil, errors.Errorf("Could not find volume snapshot name on PVC")
 	}
 
-	g := "snapshot.storage.k8s.io"
-	pvc.Spec.DataSource = &corev1api.TypedLocalObjectReference{
-		// This needs to be a pointer, since nil is used for the default group
-		APIGroup: &g,
-		Kind:     "VolumeSnapshot",
-		Name:     volumeSnapshotName,
+	for k := range annotations {
```
skriss: I'm not sure that we need L54-L62 -- can't we just iterate over `pvc.Annotations` since we've converted to a structured type? The meta accessor seems superfluous now.

velero-csi-plugin/csi_restorer.go
```diff
@@ -63,16 +66,35 @@ func (p *CSIRestorer) Execute(input *velero.RestoreItemActionExecuteInput) (*vel
 		return nil, errors.Errorf("Could not find volume snapshot name on PVC")
```
skriss: I'm guessing we don't want to return an error here, since not all PVCs are expected to have a CSI snapshot. Probably just debug/info log and return early.

velero-csi-plugin/csi_restorer.go
```diff
@@ -63,16 +66,35 @@ func (p *CSIRestorer) Execute(input *velero.RestoreItemActionExecuteInput) (*vel
...
+			Kind:     "VolumeSnapshot",
+			Name:     volumeSnapshotName,
+		},
+		AccessModes: accessModes,
+		Resources:   resources,
 	}
 
 	pvcMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&pvc)
 	if err != nil {
+		p.log.Errorf("failed to convert pvc into a map, %v", errors.WithStack(err))
```
skriss: we should just return the error, not log *and* return. You can use `errors.Wrapf(...)` if you want to add an additional message to the error.

velero-csi-plugin/volumesnapshot_restorer.go
```diff
@@ -38,22 +39,36 @@ func (p *VSRestorer) AppliesTo() (velero.ResourceSelector, error) {
...
 
-	// This section of code used to set the VolumeSnapshot.Spec.Source field to nil
-	// As of the beta, this is no longer relevant, and the code has been removed.
+	if vsFromBackup.Status == nil || vsFromBackup.Status.BoundVolumeSnapshotContentName == nil {
+		return &velero.RestoreItemActionExecuteOutput{}, errors.Errorf("unable to lookup BoundVolumeSnapshotContentName from status")
+	}
+
+	vscName := *vsFromBackup.Status.BoundVolumeSnapshotContentName
+	vs.Spec = snapshotv1beta1api.VolumeSnapshotSpec{
+		Source: snapshotv1beta1api.VolumeSnapshotSource{
```
skriss: Is this not already set in the backed-up object? Can you add a comment to clarify the intent of what we're doing here?

velero-csi-plugin/volumesnapshot_restorer.go
```diff
@@ -38,22 +39,36 @@ func (p *VSRestorer) AppliesTo() (velero.ResourceSelector, error) {
...
+			VolumeSnapshotContentName: &vscName,
+		},
+	}
 
 	vsMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&vs)
 	if err != nil {
+		p.log.Errorf("failed to convert vs into a map, %v", errors.WithStack(err))
```
skriss: log or return error, not both

velero-csi-plugin/volumesnapshotcontents_restorer.go
```diff
@@ -39,35 +40,45 @@ func (p *VSCRestorer) AppliesTo() (velero.ResourceSelector, error) {
...
+	if vscFromBackup.Status == nil || vscFromBackup.Status.SnapshotHandle == nil {
+		return &velero.RestoreItemActionExecuteOutput{}, errors.Errorf("unable to lookup snapshotHandle from status")
 	}
 
-	// Make sure we're referencing the new UID for the restored VolumeSnapshot
-	// This is likely unnecessary, but what _is_ necessary is that the old UID be cleared.
-	vsc.Spec.VolumeSnapshotRef.UID = volumeSnapshot.UID
+	vscSnapshotHandle := *vscFromBackup.Status.SnapshotHandle
+	vscSnapshotClass := *vsc.Spec.VolumeSnapshotClassName
+	vsc.Spec = snapshotv1beta1api.VolumeSnapshotContentSpec{
```
skriss: not clear on why we're needing to re-instantiate the entire spec - can you add some comments to clarify what's going on here?

velero-csi-plugin/csi_restorer.go
```diff
@@ -17,14 +17,17 @@ limitations under the License.
 package main
 
 import (
+	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
+
+	snapshotv1beta1api "github.com/kubernetes-csi/external-snapshotter/v2/pkg/apis/volumesnapshot/v1beta1"
+	core_v1 "k8s.io/api/core/v1"
```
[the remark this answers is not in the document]
ashish-amarnath: copy pasta!
Fixed now!

velero-csi-plugin/csi_restorer.go
```diff
@@ -63,16 +66,35 @@ func (p *CSIRestorer) Execute(input *velero.RestoreItemActionExecuteInput) (*vel
...
+			delete(annotations, k)
+		}
+	}
+
+	pvc.SetAnnotations(annotations)
+	sc := *pvc.Spec.StorageClassName
+	accessModes := pvc.Spec.AccessModes
+	resources := pvc.Spec.Resources
+
+	pvc.Spec = corev1api.PersistentVolumeClaimSpec{
```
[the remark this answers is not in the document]
ashish-amarnath: Re-instantiating the PVCSpec seemed cleaner than modifying the existing PVCSpec. Also, I am not sure if just setting the `DataSource` is sufficient for the volume to be restored from snapshot. Other fields being set in the spec may result in undesired controllers trying to reconcile this PVC. I saw that happen w/ the annotations. Even though the spec was using only the `dataSource`, the wrong annotations triggered the wrong controllers even though the csi driver provisioned and populated the volumes as expected. The wrongly trigged controller moved this PVC into  `ClaimLost`

velero-csi-plugin/csi_restorer.go
ashish-amarnath: leftovers from debugging.

velero-csi-plugin/csi_restorer.go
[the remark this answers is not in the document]
skriss: My expectation would be that we only need to (a) clear out the `VolumeName` field, and (b) set the `DataSource` field. Can you test that approach out?

The problem with this approach is that we may inadvertently clear out other fields that we're not intending to (for example, it looks like the `Selector` and `VolumeMode` fields are both being dropped in this implementation, and any new fields that are added to the Spec down the road would be similarly affected).

velero-csi-plugin/labels_annotations.go
```diff
@@ -0,0 +1,21 @@
+/*
+Copyright 2018, 2019 the Velero contributors.
```
skriss: just 2020 :)

velero-csi-plugin/csi_restorer.go
```diff
@@ -17,64 +17,89 @@ limitations under the License.
...
+// CSIRestorer RestoreItemAction plugin's Execute function will only be invoked on items that match the returned
+// selector. A zero-valued ResourceSelector matches all resources.
 func (p *CSIRestorer) AppliesTo() (velero.ResourceSelector, error) {
 	return velero.ResourceSelector{
 		IncludedResources: []string{"persistentvolumeclaims"},
+		//TODO: add label selector volumeSnapshotLabel
 	}, nil
 }
 
 // Execute allows the RestorePlugin to perform arbitrary logic with the item being restored,
```
skriss: I would drop "allows the RestorePlugin to perform arbitrary logic" bit and just document *this* plugin's behavior.

velero-csi-plugin/csi_snapshotter.go
```diff
@@ -38,30 +38,27 @@ import (
...
+	p.log.Info("CSISnapshotterAction AppliesTo")
 
 	return velero.ResourceSelector{
 		IncludedResources: []string{"persistentvolumeclaims"},
 	}, nil
 }
 
-// Execute allows the ItemAction to perform arbitrary logic with the item being backed up,
-// in this case, setting a custom annotation on the item being backed up.
+// Execute allows the RestorePlugin to perform arbitrary logic with the item being restored,
```
skriss: I would drop "allows the RestorePlugin to perform arbitrary logic" bit and just document *this* plugin's behavior.

velero-csi-plugin/volumesnapshot_restorer.go
```diff
@@ -31,29 +32,50 @@ type VSRestorer struct {
...
+// AppliesTo returns information about which resources the VSRestorer action should be invoked for.
+// VSRestorer RestoreItemAction plugin's Execute function will only be invoked on items that match the returned
+// selector. A zero-valued ResourceSelector matches all resources.
 func (p *VSRestorer) AppliesTo() (velero.ResourceSelector, error) {
 	return velero.ResourceSelector{
 		IncludedResources: []string{"volumesnapshots.snapshot.storage.k8s.io"},
 	}, nil
 }
 
+// Execute allows the RestorePlugin to perform arbitrary logic with the item being restored,
```
skriss: I would drop "allows the RestorePlugin to perform arbitrary logic" bit and just document *this* plugin's behavior. In this case, I think it would be helpful to be more specific about exactly what this plugin is doing -- something along the lines of "Execute modifies the VolumeSnapshot's spec to set its data source to be the relevant VolumeSnapshotContent object, rather than the original PVC the snapshot was created from, so it can be statically re-bound to the VSC on restore" (please wordsmith, just giving an example). Saying "correctly restore a CSI volumesnapshot" doesn't really give me much information.

velero-csi-plugin/volumesnapshotcontents_restorer.go
```diff
@@ -32,42 +33,60 @@ type VSCRestorer struct {
...
+// AppliesTo returns information about which resources the VSCRestorer action should be invoked for.
+// VSCRestorer RestoreItemAction plugin's Execute function will only be invoked on items that match the returned
+// selector. A zero-valued ResourceSelector matches all resources.
 func (p *VSCRestorer) AppliesTo() (velero.ResourceSelector, error) {
 	return velero.ResourceSelector{
 		IncludedResources: []string{"volumesnapshotcontent.snapshot.storage.k8s.io"},
 	}, nil
 }
 
+// Execute allows the RestorePlugin to perform arbitrary logic with the item being restored,
```
skriss: I would drop "allows the RestorePlugin to perform arbitrary logic" bit and just document *this* plugin's behavior. I would also try to be more specific about what this plugin does, along the same lines of the comment for the VolumeSnapshot restorer.

velero-csi-plugin/volumesnapshot_restorer.go
```diff
@@ -31,29 +32,50 @@ type VSRestorer struct {
...
+		return &velero.RestoreItemActionExecuteOutput{}, errors.Wrapf(err, "failed to convert input.ItemFromBackup from unstructured")
 	}
 
-	// This section of code used to set the VolumeSnapshot.Spec.Source field to nil
-	// As of the beta, this is no longer relevant, and the code has been removed.
+	if vsFromBackup.Status == nil || vsFromBackup.Status.BoundVolumeSnapshotContentName == nil {
+		return &velero.RestoreItemActionExecuteOutput{}, errors.Errorf("unable to lookup BoundVolumeSnapshotContentName from status")
+	}
+
+	vscName := *vsFromBackup.Status.BoundVolumeSnapshotContentName
```
skriss: this line seems unnecessary since we want a string pointer anyway on L69

velero-csi-plugin/volumesnapshotcontents_restorer.go
```diff
@@ -32,42 +33,60 @@ type VSCRestorer struct {
...
-	if err != nil {
-		return nil, errors.WithStack(err)
+	if vscFromBackup.Status == nil || vscFromBackup.Status.SnapshotHandle == nil {
+		return &velero.RestoreItemActionExecuteOutput{}, errors.Errorf("unable to lookup snapshotHandle from status")
 	}
 
-	// Make sure we're referencing the new UID for the restored VolumeSnapshot
-	// This is likely unnecessary, but what _is_ necessary is that the old UID be cleared.
-	vsc.Spec.VolumeSnapshotRef.UID = volumeSnapshot.UID
+	vscSnapshotHandle := *vscFromBackup.Status.SnapshotHandle
```
skriss: declaration of new vars on L64 and 65 seem unnecessary

velero-csi-plugin/volumesnapshotcontents_restorer.go
skriss: As far as I can tell, the specific things we're doing here are:
- set the `Source` field to contain the `SnapshotHandle`
- remove the old VolumeSnapshot UID from the `VolumeSnapshotRef` field

If that's the case, I think this is clearer and less error-prone if we just make those modifications, rather than re-instantiating the whole spec.

Will see what @nrb thinks here as well.

velero-csi-plugin/csi_restorer.go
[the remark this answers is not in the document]
ashish-amarnath: I agree we don't want to do that. I'll fix it!

velero-csi-plugin/volumesnapshotcontents_restorer.go
ashish-amarnath: I agree that by re-instantiating the spec, we might drop fields that may be necessary. I've fixed this to avoid re-instantiating the entire spec. 
Going to mark this as resolved. feel free to unresolve it if needed.

velero-csi-plugin/csi_restorer.go
```diff
@@ -17,64 +17,89 @@ limitations under the License.
...
+		switch k {
+		case velerov1api.BackupNameLabel, volumeSnapshotLabel:
+			// keep these annotations
+		default:
+			// remove annotations like  pv.kubernetes.io/...
+			delete(annotations, k)
+		}
+	}
+
+	pvc.SetAnnotations(annotations)
```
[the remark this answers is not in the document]
ashish-amarnath: fixed!

velero-csi-plugin/volumesnapshot_restorer.go
ashish-amarnath: I am stealing your suggestion!

velero-csi-plugin/volumesnapshot_restorer.go
ashish-amarnath: Right. I make this mistake of using C like `strdup`, bc the pointer may be freed under me, to get my own pointer instead of reusing it! Here we share the pointers 😄

velero-csi-plugin/csi_snapshotter.go
```diff
@@ -149,10 +143,17 @@ func (p *CSISnapshotter) Execute(item runtime.Unstructured, backup *velerov1api.
 	if annotations == nil {
 		annotations = make(map[string]string)
 	}
-	annotations[veleroSnapshotNameAnnotation] = upd.Name
+	annotations[volumeSnapshotLabel] = upd.Name
 	annotations[velerov1api.BackupNameLabel] = backup.Name
 	accessor.SetAnnotations(annotations)
 
+	labels := accessor.GetLabels()
```
skriss: the `accessor` is probably unnecessary here too, but not blocking, we can clean it up more later

skriss:
In https://github.com/vmware-tanzu/velero-plugin-for-csi/pull/18#issue-381823515 it says this fixes https://github.com/vmware-tanzu/velero/issues/2141, which is about restoring across clusters - is that correct, or should this be for https://github.com/vmware-tanzu/velero/issues/2140 (restoring within the same cluster)?

velero-csi-plugin/csi_snapshotter.go
ashish-amarnath: I was going to clean this up in a follow-up PR. But... 😄 
Refactored and added some tests

ashish-amarnath:
Going to take this PR through another round of testing, making sure that the refactor didn't break things.
Putting it on hold till that's done.

velero-csi-plugin/csi_snapshotter.go
ashish-amarnath: The `accessor` here is necessary as the annotations and labels need to be persisted.
We can refactor and clean up in a later PR

skriss: Yep -- like I said, not blocking, so fine to address in a separate PR.

To be clear on what I mean -- we're already converting `item` to a typed PVC object at https://github.com/vmware-tanzu/velero-plugin-for-csi/pull/18/files#diff-f8c4865ab47acc13ac1a532be0ba9e0dR66-R69, so what I would do is make any modification you need to `pvc.Annotations`, `pvc.Labels` etc, then at the bottom of this function, convert `pvc` back to an Unstructured and return it, rather than returning `item` directly.

It's not a huge deal either way, but that's more consistent with other plugins, and we're already converting `item` (Unstructured) to `pvc` (typed PVC), so may as well use `pvc`.

skriss: See, for example, https://github.com/vmware-tanzu/velero/blob/master/pkg/restore/service_action.go#L48-L68, where the flow is:

1. convert the unstructured input to a typed Service object
2. manipulate the typed Service object
3. convert the Service object back to an unstructured representation
4. return the unstructured

ashish-amarnath: will do this in a follow-up

ashish-amarnath:
This PR looked good on testing and takes care of in-cluster restore and cluster migration.

ashish-amarnath:
@nrb and @carlisia this is waiting for your review 😄

carlisia:
👀

velero-csi-plugin/csi_restorer.go
```diff
@@ -17,64 +17,79 @@ limitations under the License.
 package main
 
 import (
+	"github.com/pkg/errors"
```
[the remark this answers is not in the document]
ashish-amarnath: I missed this one. I have another PR in the pipeline. I will fix this in that PR.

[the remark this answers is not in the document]
carlisia: Ok, will merge then1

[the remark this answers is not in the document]
ashish-amarnath: fixed in https://github.com/vmware-tanzu/velero-plugin-for-csi/pull/22

PR #19, opened 2020-03-06
log a warning if the VolumeSnapshotClass's deletion policy is not Retain

ashish-amarnath:


skriss:
Need to think about this one some more. I agree that if there's an orderly deletion of the namespace in question, the storage snapshot would be lost, but if the whole cluster goes down, or if the user is trying to migrate from cluster A to cluster B, then that's not necessarily the case.

ashish-amarnath:
Yes, in the scenario of moving workloads across clusters, either migration or cluster loss, this check will not be useful. However, without this check, there will be no way for users to recover from deletion of a namespace, unless the underlying storage system can restore the snapshot.

It also would be a fair ask of the storage system to be able to restore snapshots.

That said, we can change this into a warning log message and proceed instead of failing the backup. WDYT?

velero-csi-plugin/csi_snapshotter.go
```diff
@@ -117,6 +117,14 @@ func (p *CSISnapshotter) Execute(item runtime.Unstructured, backup *velerov1api.
 	}
 	p.log.Infof("volumesnapshot class=%s", snapshotClass.Name)
 
+	// If deletetionPolicy is not Retain, then in the event of a disaster, the namespace is lost with the volumesnapshot object in it,
+	// the underlying volumesnapshotcontent and the volume snapshot in the storage provider is also deleted.
+	// In such a scenario, the backup objects will be useless as the snapshot handle itself will not be valid.
+	// For that reason, skip backup and error out
+	if snapshotClass.DeletionPolicy != snapshotv1beta1api.VolumeSnapshotContentRetain {
```
nrb: I think for at least the beta period, this should probably be a warning, and if it becomes an issue we can move towards completely disabling snapshots. Originally, I wanted to make this policy a prerequisite, but I think that can be a documentation-only prereq.

Also, updating the deletion policy is unfortunately not possibly, or wasn't in the alpha. The `VolumeSnapshotClass` objects couldn't be patched at all after creation.

ashish-amarnath: Updated to warnings instead of errors

go.sum
```diff
@@ -373,24 +378,26 @@ gopkg.in/yaml.v2 v2.2.2 h1:ZCJp+EgiOT7lHqUV2J862kp8Qj64Jo6az82+3Td9dZw=
...
 k8s.io/api v0.17.0 h1:H9d/lw+VkZKEVIUc8F3wgiQ+FUXTTr21M87jXLU7yqM=
 k8s.io/api v0.17.0/go.mod h1:npsyOePkeP0CPwyGfXDHxvypiYMJxBWAMpQxCaJ4ZxI=
-k8s.io/api v0.17.2 h1:NF1UFXcKN7/OOv1uxdRz3qfra8AHsPav5M93hlV9+Dc=
-k8s.io/api v0.17.2/go.mod h1:BS9fjjLc4CMuqfSO8vgbHPKMt5+SF0ET6u/RVDihTo4=
+k8s.io/api v0.17.3 h1:XAm3PZp3wnEdzekNkcmj/9Y1zdmQYJ1I4GKSBBZ8aG0=
+k8s.io/api v0.17.3/go.mod h1:YZ0OTkuw7ipbe305fMpIdf3GLXZKRigjtZaV5gzC2J0=
 k8s.io/apimachinery v0.17.0/go.mod h1:b9qmWdKlLuU9EBh+06BtLcSf/Mu89rWL33naRxs1uZg=
 k8s.io/apimachinery v0.17.1-beta.0 h1:0Wl/KpAiFOMe9to5h8x2Y6JnjV+BEWJiTcUk1Vx7zdE=
 k8s.io/apimachinery v0.17.1-beta.0/go.mod h1:b9qmWdKlLuU9EBh+06BtLcSf/Mu89rWL33naRxs1uZg=
-k8s.io/apimachinery v0.17.2 h1:hwDQQFbdRlpnnsR64Asdi55GyCaIP/3WQpMmbNBeWr4=
```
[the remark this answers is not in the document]
ashish-amarnath: didn't realize these come in. I reverted and and ran ci. These weren't necessary

velero-csi-plugin/csi_snapshotter.go
```diff
@@ -125,6 +125,15 @@ func (p *CSISnapshotter) Execute(item runtime.Unstructured, backup *velerov1api.
 	}
 	p.log.Infof("volumesnapshot class=%s", snapshotClass.Name)
 
+	// If deletetionPolicy is not Retain, then in the event of a disaster, the namespace is lost with the volumesnapshot object in it,
+	// the underlying volumesnapshotcontent and the volume snapshot in the storage provider is also deleted.
+	// In such a scenario, the backup objects will be useless as the snapshot handle itself will not be valid.
+	// For that reason, skip backup and error out
```
skriss: let's remove or change the comment on L131 since we're not doing that now

skriss:
other than the prev comment, LGTM

velero-csi-plugin/csi_snapshotter.go
```diff
@@ -125,6 +125,14 @@ func (p *CSISnapshotter) Execute(item runtime.Unstructured, backup *velerov1api.
 	}
 	p.log.Infof("volumesnapshot class=%s", snapshotClass.Name)
 
+	// If deletetionPolicy is not Retain, then in the event of a disaster, the namespace is lost with the volumesnapshot object in it,
+	// the underlying volumesnapshotcontent and the volume snapshot in the storage provider is also deleted.
+	// In such a scenario, the backup objects will be useless as the snapshot handle itself will not be valid.
+	if snapshotClass.DeletionPolicy != snapshotv1beta1api.VolumeSnapshotContentRetain {
+		p.log.Warnf("DeletionPolicy on VolumeSnapshotClass %s is not %s; Deletion of VolumeSnapshot objects will lead to deletion of snapshot in the storage provider; ",
+			snapshotClass.Name, snapshotv1beta1api.VolumeSnapshotContentRetain, snapshotClass.DeletionPolicy, snapshotv1beta1api.VolumeSnapshotContentRetain)
+		p.log.Warnf("Please consider update DeletionPolicy from %s to %s", snapshotClass.DeletionPolicy, snapshotv1beta1api.VolumeSnapshotContentRetain)
```
nrb: `update` -> `updating`, though the VolumeSnapshotClasses are immutable, so the user basically has to migrate their data to  a new volume.

velero-csi-plugin/csi_snapshotter.go
```diff
@@ -125,6 +125,13 @@ func (p *CSISnapshotter) Execute(item runtime.Unstructured, backup *velerov1api.
 	}
 	p.log.Infof("volumesnapshot class=%s", snapshotClass.Name)
 
+	// If deletetionPolicy is not Retain, then in the event of a disaster, the namespace is lost with the volumesnapshot object in it,
+	// the underlying volumesnapshotcontent and the volume snapshot in the storage provider is also deleted.
+	// In such a scenario, the backup objects will be useless as the snapshot handle itself will not be valid.
+	if snapshotClass.DeletionPolicy != snapshotv1beta1api.VolumeSnapshotContentRetain {
+		p.log.Warnf("DeletionPolicy on VolumeSnapshotClass %s is not %s; Deletion of VolumeSnapshot objects will lead to deletion of snapshot in the storage provider; ",
```
skriss: since you removed the subsequent log line, please end this one with just a `.` instead of a `; `

PR #24, opened 2020-03-17
🏃‍♀️Use stretchr

ashish-amarnath:

### DIFF
--- a/velero-csi-plugin/csi_restorer_test.go
+++ b/velero-csi-plugin/csi_restorer_test.go
@@ -19,6 +19,8 @@ package main
 import (
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+
 	corev1api "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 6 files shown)


velero-csi-plugin/csi_restorer_test.go
```diff
@@ -96,26 +98,21 @@ func TestResetPVCAnnotations(t *testing.T) {
...
-			if tc.pvc.Annotations == nil {
-				t.Errorf("%s failed, unexpected PVC annotations returned, Want: non-nil; Got: nil", tc.name)
-			}
-			if tc.expectNonEmptyAnnotations && len(tc.pvc.Annotations) != 0 {
-				t.Errorf("%s failed, unexpected count of PVC annotations returned, Want: 0; Got: %d", tc.name, len(tc.pvc.Annotations))
+
+			assert.NotNil(t, tc.pvc.Annotations, "unexpected PVC annotations returned, Want: non-nil; Got: nil")
+
+			if tc.expectNonEmptyAnnotations {
+				assert.Emptyf(t, tc.pvc.Annotations, "unexpected count of PVC annotations returned, Want: 0; Got: %d", len(tc.pvc.Annotations))
```
skriss: something seems reversed here - when tc.expect**Non**EmptyAnnotations is true, we assert that tc.pvc.Annotations is empty?

ashish-amarnath: That was confusing. I updated the tests to make it more clear. PTAL.

velero-csi-plugin/csi_restorer_test.go
```diff
@@ -90,33 +99,31 @@ func TestResetPVCAnnotations(t *testing.T) {
...
 			resetPVCAnnotations(&tc.pvc, tc.preserveAnnotations)
-			if tc.pvc.Annotations == nil {
-				t.Errorf("%s failed, unexpected PVC annotations returned, Want: non-nil; Got: nil", tc.name)
-			}
-			if tc.expectNonEmptyAnnotations && len(tc.pvc.Annotations) != 0 {
-				t.Errorf("%s failed, unexpected count of PVC annotations returned, Want: 0; Got: %d", tc.name, len(tc.pvc.Annotations))
-			}
-
-			if tc.expectNonEmptyAnnotations {
+			if tc.expectEmptyAnnotations {
```
skriss: OK, so now that I'm looking at this more closely, I think we can remove L115-126 and replace it with `assert.Equal(t, tc.expectedAnnotations, tc.pvc.Annotations)`, right? 

If so, then can also remove the `expectEmptyAnnotations` field from the anonymous test case struct.

ashish-amarnath: Lol! Yes. That would def simplify this!
Updated the tests. Thanks for catching this!

PR #25, opened 2020-03-19
wait for VolumeSnapshotContents to have a snapshot handle during backup

skriss:
Signed-off-by: Steve Kriss <krisss@vmware.com>

Hopefully the comment explains why this is needed. For now I just reused the existing "wait" pattern from that func, though we'll need to add timeouts and maybe make some other changes to handle snapshot failures later too.


skriss:
fixed unit test.

skriss:
@nrb need another approve here since i had to push a unit test fix, thanks!

PR #26, opened 2020-03-19
misc cleanup

skriss:
Couple misc cleanup items, broken out by commit


ashish-amarnath:
@skriss Curious, did you use the examples from the repo for your tests? I haven't used them and not sure what state they are in. But I am all in w/ relocating them and fixing it in their new spot.

skriss:
I have not actually used them, I thought you had added them but I was using the samples from the Azure repo :)

I think it would be useful to have a simple example setup though, so I'd vote to keep them and fix anything we need to.

ashish-amarnath:
I will update those examples.

PR #35, opened 2020-04-03
Await snap content reconciliation only for velero created volumeSnapshots

ashish-amarnath:
Signed-off-by: Ashish Amarnath ashisham@vmware.com

This branch was cut from the branch for #34 

Changes in this PR:

- modify csi_snapshotter BackupItemAction to:
    - only return VolumeSnapshot as additional items to backup
    - not return VolumesnapshotClass and VolumeSnapshotContents as additional items to backup
- modify volumesnapshot_action BackupItemAction to:
    - wait for VolumeSnapshotContents reconciliation only for VolumeSnapshots created by Velero, while backing up a CSI backed PVC


internal/backup/volumesnapshot_action.go
```diff
@@ -60,8 +60,18 @@ func (p *VolumeSnapshotBackupItemAction) Execute(item runtime.Unstructured, back
...
 
+	// determine if we are backing up a volumesnapshot that was created by velero while performing backup of a
+	// CSI backed PVC.
+	// For volumesnapshots that were created during the backup of a CSI backed PVC, we will wait for the volumecontents to
+	// be available.
+	// For volumesnapshots created outside of velero, we expect the volumesnapshotcontent to be available prior to backing up
+	// the volumesnapshot. In case of a failure, backup should be re-attempted after the CSI driver has reconciled the volumesnapshot.
+	// existence of the velerov1api.BackupNameLabel indicates that the volumesnapshot was created while backing up a
+	// CSI backed PVC.
+	_, toWait := vs.Labels[velerov1api.BackupNameLabel]
```
skriss, line 71: I think we also want to check if the snapshot was created for **this** Velero backup, not just any backup. So need to compare the label value to `backup.Name`. 

Unfortunately, backup name isn't foolproof here, since it's not unique over time (backup `foo` could be created, deleted, and recreated). We may need to use the backup UID, which will be unique over time. Need to think about this a little more, since using UID comes with its own set of gotchas.

For now, let's at least check the name and add a `// TODO `.

skriss, line 71: maybe `shouldWait` instead of `toWait`?

internal/util/util.go
```diff
@@ -135,17 +135,30 @@ func GetVolumeSnapshotClassForStorageClass(provisioner string, snapshotClient sn
 }
 
 // GetVolumeSnapshotContentForVolumeSnapshot returns the volumesnapshotcontent object associated with the volumesnapshot
-func GetVolumeSnapshotContentForVolumeSnapshot(volSnap *snapshotv1beta1api.VolumeSnapshot, snapshotClient snapshotter.SnapshotV1beta1Interface, log logrus.FieldLogger) (*snapshotv1beta1api.VolumeSnapshotContent, error) {
+func GetVolumeSnapshotContentForVolumeSnapshot(volSnap *snapshotv1beta1api.VolumeSnapshot, snapshotClient snapshotter.SnapshotV1beta1Interface, log logrus.FieldLogger, wait bool) (*snapshotv1beta1api.VolumeSnapshotContent, error) {
```
skriss, line 138: I'll look at this in some more detail in the morning, but I think we may be able to simplify by using some of the helpers from `k8s.io/apimachinery/pkg/util/wait`.

internal/backup/volumesnapshot_action.go
ashish-amarnath: fixed

ashish-amarnath: done

internal/util/util.go
ashish-amarnath: I'll also look for something we can use from there.

internal/backup/volumesnapshot_action.go
```diff
@@ -60,8 +60,24 @@ func (p *VolumeSnapshotBackupItemAction) Execute(item runtime.Unstructured, back
...
+	// the volumesnapshot. In case of a failure, backup should be re-attempted after the CSI driver has reconciled the volumesnapshot.
+	// existence of the velerov1api.BackupNameLabel indicates that the volumesnapshot was created while backing up a
+	// CSI backed PVC.
+	backupName, exists := vs.Labels[velerov1api.BackupNameLabel]
+	// TODO: use UID of the backup object instead of just looking for a match in the backup name. This will ensure we don't wait in cases where
+	// backup names are reused.
+	// We want to await reconciliation of only those volumesnapshots created during the ongoing backup.
+	// For this we will wait only if the backup label exists on the volumesnapshot object and the
+	// backup name is the same as that of the value of the backupLabel
+	shouldWait := exists && (backupName == backup.Name)
```
skriss, line 77: there's no need to look at `exists` since the map value will come back as an empty string if it doesn't exist - you can simplify L71&77 as:

`shouldWait := vs.Labels[velerov1aapi.BackupNameLabel] == backup.Name`

internal/util/util.go
skriss: `wait.PollImmediate` would probably be a good fit

internal/util/util.go
```diff
@@ -135,54 +136,51 @@ func GetVolumeSnapshotClassForStorageClass(provisioner string, snapshotClient sn
 }
 
 // GetVolumeSnapshotContentForVolumeSnapshot returns the volumesnapshotcontent object associated with the volumesnapshot
-func GetVolumeSnapshotContentForVolumeSnapshot(volSnap *snapshotv1beta1api.VolumeSnapshot, snapshotClient snapshotter.SnapshotV1beta1Interface, log logrus.FieldLogger, wait bool) (*snapshotv1beta1api.VolumeSnapshotContent, error) {
+func GetVolumeSnapshotContentForVolumeSnapshot(volSnap *snapshotv1beta1api.VolumeSnapshot, snapshotClient snapshotter.SnapshotV1beta1Interface, log logrus.FieldLogger, shouldWait bool) (*snapshotv1beta1api.VolumeSnapshotContent, error) {
```
skriss, line 139: I think I would handle the `shouldWait == false` case on its own at the top of this function (outside the loop), for a few reasons:
- makes the polling simpler since you don't have to deal with that case
- I don't see a need to query the API again for the `VolumeSnapshot` - we can just use the version that's been passed into the plugin
- if it doesn't have a `.Status.BoundVolumeSnapshotContentName`, or that VolumeSnapshotContent doesn't have a `snapshotHandle` that shouldn't be an error since we're not waiting for them to provision, **which is different** than for Velero-created VolumeSnapshots

Something like the following:

```go
if !shouldWait {
    if volSnap.Status == nil || volSnap.Status.BoundVolumeSnapshotContentName == nil {
        // no bound VSC: it's not an error, but there's nothing additional to restore
        return nil, nil
    }

    vsc, err := snapshotClient.VolumeSnapshotContents().Get(*volSnap.Status.BoundVolumeSnapshotContentName, metav1.GetOptions{})
    if err != nil {
        return nil, errors.Wrap(err, "error getting volume snapshot content from API")
    }

    return vsc, nil
}
```

Note, you'll also have to handle nil `VolumeSnapshotContent` and nil `VolumeSnapshotContent.Status.SnapshotHandle` in the calling function if you use this approach.

Then, you can remove having to worry about that case from the polling below.

internal/util/util.go
```diff
@@ -135,36 +136,49 @@ func GetVolumeSnapshotClassForStorageClass(provisioner string, snapshotClient sn
...
+				log.Infof("Waiting for volumesnapshotcontents %s to have snapshot handle. Retrying in %ds", snapshotContent.Name, interval/time.Second)
+			}
+			return false, nil
 		}
 
-		break
+		return true, nil
+	})
+
+	if err != nil {
```
skriss, line 180: you might want to differentiate between a `wait.ErrWaitTimeout` and any other error here, and wrap them differently

internal/backup/volumesnapshot_action.go
```diff
@@ -60,8 +60,24 @@ func (p *VolumeSnapshotBackupItemAction) Execute(item runtime.Unstructured, back
...
+	// CSI backed PVC.
+	// For volumesnapshots that were created during the backup of a CSI backed PVC, we will wait for the volumecontents to
+	// be available.
+	// For volumesnapshots created outside of velero, we expect the volumesnapshotcontent to be available prior to backing up
+	// the volumesnapshot. In case of a failure, backup should be re-attempted after the CSI driver has reconciled the volumesnapshot.
+	// existence of the velerov1api.BackupNameLabel indicates that the volumesnapshot was created while backing up a
+	// CSI backed PVC.
+
+	// TODO: use UID of the backup object instead of just looking for a match in the backup name. This will ensure we don't wait in cases where
+	// backup names are reused.
```
nrb, line 73: Once we have proper garbage collection, I don't think the re-use case should matter, should it?

When deleting the `Backup`, all namespaced `VolumeSnapshot`s should also be deleted, though any `VolumeSnapshotContents` objects with a deletion policy of `Retain` won't be able to be touched.

So we can't re-use the `Backup` name while the object exists, and deleting the `Backup` manually via `kubectl` is unsupported, requiring manual clean up....or we have our GC look at `VolumeSnapshots` pointing to a `Backup` that doesn't exist and delete them for the user.

internal/util/util.go
ashish-amarnath: I used your suggestion. But, IMO, I don't think it made the polling logic any simpler.
The only part in the polling that was waiting aware was a log message.

internal/backup/volumesnapshot_action.go
ashish-amarnath: Yes, once we have proper garbage collection. This shouldn't matter.
We can remove the UID `//TODO` when we've implemented the backup deletion.

internal/util/util.go
```diff
@@ -135,36 +136,56 @@ func GetVolumeSnapshotClassForStorageClass(provisioner string, snapshotClient sn
...
+		}
+		vsc, err := snapshotClient.VolumeSnapshotContents().Get(*volSnap.Status.BoundVolumeSnapshotContentName, metav1.GetOptions{})
+		if err != nil {
+			return nil, errors.Wrap(err, "error getting volume snapshot content from API")
+		}
+		return vsc, nil
+	}
+
+	// We'll wait 10m for the VSC to be reconciled polling every 5s
+	timeout := 10 * time.Minute
```
skriss, line 153: What's our expectation for how long this will actually take? It should be pretty quick, right, since we don't have to wait for the snapshot to be "ready to use"?

I have a feeling we're going to need to make this configurable eventually - maybe just add a TODO for now.

ashish-amarnath: yes, this should be pretty quick. Esp, now that we are trying to get the volumesnapshotcontents in the volumesnapshot backup plugin.

internal/backup/volumesnapshot_action.go
```diff
@@ -60,37 +60,62 @@ func (p *VolumeSnapshotBackupItemAction) Execute(item runtime.Unstructured, back
...
+			GroupResource: kuberesource.VolumeSnapshotContents,
+			Name:          vsc.Name,
+		})
 
-	// save newly applied annotations into the backed-up volumesnapshot item
-	util.AddAnnotations(&vs.ObjectMeta, vals)
+		// Capture storage provider snapshot handle and CSI driver name
+		// to be used on restore to create a static volumesnapshotcontent that will be the source of the volumesnapshot.
+		vals := map[string]string{
+			util.VolumeSnapshotHandleAnnotation: *vsc.Status.SnapshotHandle,
```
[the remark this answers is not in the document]
ashish-amarnath: thanks for catching that!

internal/backup/volumesnapshot_action.go
```diff
@@ -60,37 +60,62 @@ func (p *VolumeSnapshotBackupItemAction) Execute(item runtime.Unstructured, back
...
+	// backup names are reused.
+	// We want to await reconciliation of only those volumesnapshots created during the ongoing backup.
+	// For this we will wait only if the backup label exists on the volumesnapshot object and the
+	// backup name is the same as that of the value of the backupLabel
+	shouldWait := vs.Labels[velerov1api.BackupNameLabel] == backup.Name
+
 	p.Log.Infof("Getting VolumesnapshotContent for Volumesnapshot %s/%s", vs.Namespace, vs.Name)
-	vsc, err := util.GetVolumeSnapshotContentForVolumeSnapshot(&vs, snapshotClient.SnapshotV1beta1(), p.Log)
+
+	additionalItems := []velero.ResourceIdentifier{
```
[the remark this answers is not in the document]
ashish-amarnath: agreed! moved.

PR #36, opened 2020-04-04
refactor and use func to add annotations and labels

ashish-amarnath:

Branch for this PR was cut from that of #35 

Changes in this PR:

- add a new util function to add labels on objects
- use the util functions to add annotations and labels, during backup, on a PVC backed by CSI


internal/util/util.go
```diff
@@ -236,13 +236,21 @@ func IsVolumeSnapshotHasVSCDeleteSecret(vs *snapshotv1beta1api.VolumeSnapshot) b
...
-
 	for k, v := range vals {
-		o.Annotations[k] = v
+		(*cur)[k] = v
 	}
 }
+
+// AddAnnotations adds the supplied key-values to the annotations on the object
+func AddAnnotations(o *metav1.ObjectMeta, vals map[string]string) {
+	addToMap(&o.Annotations, vals)
```
skriss, line 250: I would probably drop the `addToMap` helper and just duplicate the relevant code in both `AddAnnotations` and `AddLabels` - I get why you're doing the pointer stuff with the maps but I think it's probably simpler to avoid it and just duplicate the few lines of code.

skriss:
LGTM once #35 is merged..

internal/util/util_test.go
```diff
@@ -1291,3 +1291,66 @@ func TestSetAnnotations(t *testing.T) {
 		})
 	}
 }
+
+func TestAddLables(t *testing.T) {
```
nrb, line 1295: `TestAddLabels`

ashish-amarnath: Good catch!

PR #47, opened 2020-04-27
add velero backup label for vsc created in vs backup item action

### LINKED ISSUE #46: Dynamically created volumesnapshotcontents should be tagged with the backup label.
In the following scenario:
1. Take a backup of the workload:
    - create volumesnapshot and volumesnapshotcontent (dynamically created with deletion policy retain)
2. Delete workload namespace:
    - deletes the volumesnapshot object, but preserves the volumesnapshotcontent.

Now deleting the backup: This will not cleanup the volumesnapshotcontents will not get cleaned up in the storage provider.
Tagging the dynamically created volumesnapshotcontent with the backup label makes it discoverable to perform deletion.

cc @skriss @nrb

ashish-amarnath:

Fixes #46

### DIFF
--- a/internal/backup/volumesnapshot_action.go
+++ b/internal/backup/volumesnapshot_action.go
@@ -17,12 +17,15 @@ limitations under the License.
 package backup
 
 import (
+	"fmt"
+
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 
 	snapshotv1beta1api "github.com/kubernetes-csi/external-snapshotter/v2/pkg/apis/volumesnapshot/v1beta1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/vmware-tanzu/velero-plugin-for-csi/internal/util"
 	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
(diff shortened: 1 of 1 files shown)


internal/backup/volumesnapshot_action.go
```diff
@@ -113,6 +116,26 @@ func (p *VolumeSnapshotBackupItemAction) Execute(item runtime.Unstructured, back
...
+			// However, in the event that the Volumesnapshot object is deleted outside of the backup deletion process, it is possible that
+			// the dynamically created volumesnapshotcontent object will be left as an orphaned and non-discoverable resource in the cluster as well
+			// as in the storage provider. To avoid piling up of such orphaned resources, we will want to discover and delete the dynamically created
+			// volumesnapshotcontents. We do that by adding the "velero.io/backup-name" label on the volumesnapshotcontent.
+			// Further, we want to add this label only on volumesnapshotcontents that were created during an ongoing velero backup.
+
+			pb := []byte(fmt.Sprintf(`{"metadata":{"labels":{"%s":"%s"}}}`, velerov1api.BackupNameLabel, backup.Name))
+			_, vscPatchError = snapshotClient.SnapshotV1beta1().VolumeSnapshotContents().Patch(vsc.Name, types.MergePatchType, pb)
+			if vscPatchError != nil {
+				p.Log.Errorf("Failed to patch volumesnapshotcontent %s: %v", vsc.Name, vscPatchError)
```
skriss, line 136: I am wondering if we should log this as a warning, and not return the error, since it's not critical for the purposes of creating/restoring the backup.

If we return this error (below), then the backup of the PVC will be aborted/it will not be captured in the tarball, which probably *isn't* what we want here.

ashish-amarnath: I thought returning this error would result in partial failure of the backup. Meaning, the backup was performed, but there were errors.
If returning an error is going to abort the backup, then that is not what we want. Going to change the log message to a warning and not return this error.

skriss: it doesn't abort the entire backup, but it does skip backing up the "current item" (i.e. the PVC in this case). See https://github.com/vmware-tanzu/velero/blob/master/pkg/backup/item_backupper.go#L213-L224

ashish-amarnath: Thanks for pointing that out! I've updated to log a warning and not return the error.

PR #48, opened 2020-05-06
Use GetValidName for label values

ashish-amarnath:

### DIFF
--- a/internal/backup/pvc_action.go
+++ b/internal/backup/pvc_action.go
@@ -32,6 +32,7 @@ import (
 	"github.com/vmware-tanzu/velero-plugin-for-csi/internal/util"
 	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	"github.com/vmware-tanzu/velero/pkg/kuberesource"
+	"github.com/vmware-tanzu/velero/pkg/label"
 	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
 	"github.com/vmware-tanzu/velero/pkg/util/boolptr"
 )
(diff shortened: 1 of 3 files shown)


internal/backup/pvc_action.go
```diff
@@ -122,7 +123,7 @@ func (p *PVCBackupItemAction) Execute(item runtime.Unstructured, backup *velerov
 			GenerateName: "velero-" + pvc.Name + "-",
 			Namespace:    pvc.Namespace,
 			Labels: map[string]string{
-				velerov1api.BackupNameLabel: backup.Name,
+				velerov1api.BackupNameLabel: label.GetValidName(backup.Name),
 			},
 		},
 		Spec: snapshotv1beta1api.VolumeSnapshotSpec{
```
nrb, line 129: Lines 144 and 145 need it, too.

This is tricky because we'll be sanitizing names that we rely on being accurate in the API server :(

ashish-amarnath: Those get addressed in the `util.AddLabels` function

skriss:
this all looks good! I didn't double-check to see if there are any other spots that need to be updated though

ashish-amarnath:
I think we want to do the same for annotations as well.

________________________________

skriss:
It's not really necessary for annotations, since they can contain arbitrary values - is there a specific reason to do it for them? With labels, we have [this specific set of constraints](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set).

carlisia:
LGTM, waiting to see if more will be added. 👁️

ashish-amarnath:
We store the snapshotHandle and CSI driver name as metadata in the annotations. I don't think those should be modified. Leaving the annotations unchanged.

PR #50, opened 2020-05-08
Setting up gh actions for image build and push

ashish-amarnath:

~Let's see if this works. 🤞~

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,39 @@
+name: Push
+
+on:
+  push:
+    branches: [ master ]
+    tags:
+      - '*'
+
+jobs:
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+
+    - name: Set up Go 1.14
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.14
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Build
+      run: make all
+
+    - name: Test
+      run: make test
+    
+    - name: Get Tag
+      id: get_tag
+      run: |
+        echo ::set-output name=TAG::$(echo $GITHUB_REF | cut -d / -f 3)
+        
+    - name: Publish container image
+      run: |
+        docker login -u ${{ secrets.DOCKER_USER }} -p ${{ secrets.DOCKER_PASSWORD }}
+        TAG=${{ steps.get_tag.outputs.TAG }} make push
--- a/Makefile
+++ b/Makefile
@@ -33,11 +33,13 @@ platform_temp = $(subst -, ,$(ARCH))
 GOOS = $(word 1, $(platform_temp))
 GOARCH = $(word 2, $(platform_temp))
 
+.PHONY: all
 all: $(addprefix build-, $(BIN))
 
 build-%:
 	$(MAKE) --no-print-directory BIN=$* build
 
+.PHONY: local
 local: build-dirs
 	GOOS=$(GOOS) \
 	GOARCH=$(GOARCH) \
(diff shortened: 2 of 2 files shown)


ashish-amarnath:
disabling the mandatory 2 reviews to enable faster turn around for testing this PR.

skriss:
will review shortly, but one initial comment is that we'll want to build & push images when a tag is pushed as well - so when we push the `v0.1.0` tag, we get a docker image tagged with that version.

ashish-amarnath:
Example Action runs:
1. Push to master: https://github.com/ashish-amarnath/velero-plugin-for-csi/runs/657265815?check_suite_focus=true
2. Pushing a tag: https://github.com/ashish-amarnath/velero-plugin-for-csi/runs/657279864?check_suite_focus=true

Please look under the: ` Publish container image` job for the the tag associated with the container image.

More runs:
https://github.com/ashish-amarnath/velero-plugin-for-csi/actions

.github/workflows/go.yml
```diff
@@ -0,0 +1,46 @@
...
+        go-version: 1.14
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Get dependencies
+      run: |
+        go get -v -t -d ./...
+        if [ -f Gopkg.toml ]; then
```
skriss, line 28: looks like we don't need L28-31 since we're not using `dep`. Actually, do we even need this "Get dependencies" step? Modules should be pulled automatically during the build, right?

.github/workflows/go.yml
```diff
@@ -0,0 +1,46 @@
...
+      run: make test
+    
+    - name: Get Tag
+      id: get_tag
+      run: |
+        echo ::set-output name=TAG::$(echo $GITHUB_REF | cut -d / -f 3)
+        
+    - name: Publish container image
+      run: |
+        TAG=${{ steps.get_tag.outputs.TAG }} make container
```
skriss, line 46: this should be `make push`, right?

Also, need to figure out how to provide the docker hub credentials.

ashish-amarnath: Yes. That has to be a `push`. Added that target to the Makefile. Wanted to get this to work before pushing random images to the registry from my fork.

.github/workflows/go.yml
ashish-amarnath: Yes, it shouldn't be necessary. the GH template gave it to me and I didn't remove it.
Will remove it.

.github/workflows/go.yml
```diff
@@ -0,0 +1,43 @@
...
+      run: make test
+    
+    - name: Get Tag
+      id: get_tag
+      run: |
+        echo ::set-output name=TAG::$(echo $GITHUB_REF | cut -d / -f 3)
+        
+    - name: Publish container image
+      run: |
+        docker login -u ${DOCKER_USER} -p ${DOCKER_PASSWORD}
```
skriss, line 42: Do you need to explicitly populate the env vars, per [this doc](https://help.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#example-using-bash)? It's not totally clear on whether these are automatically populated from the repo's secrets or not.

ashish-amarnath: probably not. let me try that out.

ashish-amarnath: OK that works. Removed the env vars

ashish-amarnath:
Here is the example run of the actions:
1. [Building and pushing images on push to master](https://github.com/ashish-amarnath/velero-plugin-for-csi/runs/657689344?check_suite_focus=true)
2. [Building and pushing images on tag push](https://github.com/ashish-amarnath/velero-plugin-for-csi/runs/657706112?check_suite_focus=true)

Look under the `Publish container image` step

.github/workflows/go.yml
```diff
@@ -0,0 +1,43 @@
...
+    - name: Set up Go 1.14
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.14
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Get dependencies
```
skriss, line 25: I still think this step can be entirely removed, but correct me if I'm wrong :)

skriss:
the rest of this looks good to me, pending getting the secret set up.

One other thing we probably want to do is enable `latest` tagging - in our other repos, we push an image with the `latest` tag (as well as the semver tag) if it's the most recent stable release. We should probably do the same here to be consistent, and we'll have to figure it out for the other repos anyway. We can tackle it in a separate PR though.

.github/workflows/go.yml
ashish-amarnath: That wasn't required. removed.

ashish-amarnath:
New bot user `velerobot2` has been created and its creds are in GH secret. This PR should be good to go.

skriss:
LGTM.

@ashish-amarnath can you also add a follow-up issue to add `latest` tagging?

ashish-amarnath:
Created https://github.com/vmware-tanzu/velero-plugin-for-csi/issues/52

skriss:
@carlisia @nrb PTAL, thanks!

PR #54, opened 2020-05-15
🏃‍♂️ Add restore label on VSC created for static binding

ashish-amarnath:

### DIFF
--- a/internal/restore/volumesnapshot_action.go
+++ b/internal/restore/volumesnapshot_action.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/vmware-tanzu/velero-plugin-for-csi/internal/util"
 	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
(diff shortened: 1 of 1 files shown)


internal/restore/volumesnapshot_action.go
```diff
@@ -85,6 +86,9 @@ func (p *VolumeSnapshotRestoreItemAction) Execute(input *velero.RestoreItemActio
 			Annotations: map[string]string{
 				velerov1api.RestoreNameLabel: input.Restore.Name,
 			},
+			Labels: map[string]string{
```
skriss, line 89: do we need it as an annotation as well? (lines just above) If we're trying to match what Velero does, it only adds a label. Note that it also adds the backup name label, we might want to do that too.

ashish-amarnath: Removing the annotation should be OK.

I don't think we want to add the backup name label because we use that to identify resources to delete when backups expire.

PR #84, opened 2021-03-10
Only publish the container image from the root repo

jenting:
Only try to publish the container image from the root repo `vmware-tanzu/velero-plugin-for-csi`.
The forked repo doesn't have permission to do so and will always get failures.


ashish-amarnath (approved):

dsu-igeek (approved):
Thanks!

PR #89, opened 2021-06-22
Only remove annotations related to PV binding

tomob:
PVCs can have any number annotations used for any purposes. Removing
them all can brake clients' workflows.

This commit changes the removal to only remove the annotations that are
used by K8s itself to manage PVC-PV binding. All other annotations are
left intact.

Fixes [issue #3925](https://github.com/vmware-tanzu/velero/issues/3925)

Signed-off-by: Tomasz Baranski <tbaransk@redhat.com>

### DIFF
--- a/changelogs/unreleased/89-tomob
+++ b/changelogs/unreleased/89-tomob
@@ -1,0 +1,1 @@
+Only remove internal k8s annotations related to PVC-PV bindings and provisioning
--- a/internal/restore/pvc_action.go
+++ b/internal/restore/pvc_action.go
@@ -31,8 +31,14 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/vmware-tanzu/velero-plugin-for-csi/internal/util"
-	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
+)
+
+const (
+	AnnBindCompleted      = "pv.kubernetes.io/bind-completed"
+	AnnBoundByController  = "pv.kubernetes.io/bound-by-controller"
+	AnnStorageProvisioner = "volume.beta.kubernetes.io/storage-provisioner"
+	AnnSelectedNode       = "volume.kubernetes.io/selected-node"
 )
 
 // PVCRestoreItemAction is a restore item action plugin for Velero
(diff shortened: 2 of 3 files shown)


tomob:
@sseago I'd appreciate you taking a look.

internal/restore/pvc_action.go
```diff
@@ -31,10 +31,14 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 
 	"github.com/vmware-tanzu/velero-plugin-for-csi/internal/util"
-	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
 )
 
+const (
+	AnnBindCompleted     = "pv.kubernetes.io/bind-completed"
+	AnnBoundByController = "pv.kubernetes.io/bound-by-controller"
```
sseago, line 39: Looking in my current OpenShift cluster, I see that there are actually two other cluster-generated annotations on my PVCs. I don't know that it matters all that much whether they are removed or retained, but to minimize the change from this PR (before the PR, everything but velero annotations is removed, after the PR, only annotations in this list are removed), it might make sense to add these other two to the remove list as well:
```
               "volume.beta.kubernetes.io/storage-provisioner"
               "volume.kubernetes.io/selected-node"
```

sseago (commented):
I think this change makes sense overall. I'm thinking that we want to add those other two annotations I'm seeing in my cluster to the "remove" list, but if other reviewers feel there's no need to do that, then everything could be left as-is here.

internal/restore/pvc_action.go
tomob: Commit updated.

jenting (commented):
Is there any GitHub issue related to this?
It'd be good to have a change log for this PR.

tomob:
> Is there any GitHub issue related to this?
> It'd be good to have a change log for this PR.

I don't think so. Let me create one.

jenting:
> > Is there any GitHub issue related to this?
> > It'd be good to have a change log for this PR.
> 
> I don't think so. Let me create one.

Would you mind add a change log for this PR?

tomob:
> Would you mind add a change log for this PR?

Done.

changelogs/unreleased/89-tomob
```diff
@@ -0,0 +1 @@
+Only remove internal k8s annotations related to PVC-PV bindings and provisioning (fixes [#3925](https://github.com/vmware-tanzu/velero/issues/3925)).
```
jenting, line 1:
```suggestion
Only remove internal k8s annotations related to PVC-PV bindings and provisioning
```

tomob: Done

sseago (approved):

jenting (approved):
Thank you.

tomob:
Thank you!

PR #102, opened 2022-04-13
feature annotation for snapshotclass provisioner

big-appled:
## Feature

Add a new annotation in snapshotclass to let user input the provisioner name, as some CSI providers have different provisioner name and driver name.


reasonerjt:
> In one cloud provider, the csi snapclass driver name is disk.csi.everest.io, but the storageclass provisioner name is everest-csi-provisioner.

Does it violate the CSI spec?  If it's not following the spec, I don't think we should patch velero to support it.

big-appled:
> > In one cloud provider, the csi snapclass driver name is disk.csi.everest.io, but the storageclass provisioner name is everest-csi-provisioner.
> 
> Does it follow the CSI spec? If it's not following the spec, I don't think we should patch velero to support it.

K8s CSI spec does NOT require the two names (provisioner and driver) be the same. So I believe it is a valid case.

sseago (approved):

jerry-jibu:
> > In one cloud provider, the csi snapclass driver name is disk.csi.everest.io, but the storageclass provisioner name is everest-csi-provisioner.
> 
> Does it follow the CSI spec? If it's not following the spec, I don't think we should patch velero to support it.

CSI Spec doesn't define this part that snapshot driver name must be same to corresponding storage class provisioner name for same storage.  I think it's a kind of common convention that most vendors follow it with above corner case from huawei cloud.

reasonerjt:
@jerry-jibu sorry for the late response, as we synced offline with jibu folks, this is a corner case due to impl of Huawei's CSI driver, I suggest we hold this PR until there are more of such use cases.

reasonerjt (changes requested):
Blocking the merge of this PR, until we reach consensus

jkroepke:
Hi @reasonerjt! I would also like to share an other use case, maybe its an different use case.

In our setup, all our snapshots requires tags to identity the source of the snapshots. By default, the CSI Snapshot driver to not create any tags.

In conclusion, we need a dedicated SnapshotClass for each Namespace, like:

```
apiVersion: snapshot.storage.k8s.io/v1beta1
kind: VolumeSnapshotClass
metadata:
  name: csi-azuredisk-vsc-inc-1000100
driver: disk.csi.azure.com
deletionPolicy: Delete
parameters:
  tags: 'mandantID=1000100'
```

In conclusion, we have multiple VolumeSnapshotClasses for one CSI StorageClass. Currently, I'm not able to define the snapshop class in velero.

PR #108, opened 2022-06-07
Update VSC's DeletionPolicy to Retain when it is set to Delete.

blackpiglet:
Fix issue https://github.com/vmware-tanzu/velero/issues/4945

Signed-off-by: Xun Jiang <jxun@vmware.com>

### DIFF
--- a/internal/restore/volumesnapshot_action.go
+++ b/internal/restore/volumesnapshot_action.go
@@ -54,6 +54,10 @@ func resetVolumeSnapshotSpecForRestore(vs *snapshotv1api.VolumeSnapshot, vscName
 	vs.Spec.Source.VolumeSnapshotContentName = vscName
 }
 
+func resetVolumeSnapshotAnnotation(vs *snapshotv1api.VolumeSnapshot) {
+	vs.ObjectMeta.Annotations[util.CSIVSCDeletionPolicy] = string(snapshotv1api.VolumeSnapshotContentRetain)
+}
+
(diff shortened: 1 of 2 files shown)


internal/restore/volumesnapshot_action.go
```diff
@@ -93,6 +93,12 @@ func (p *VolumeSnapshotRestoreItemAction) Execute(input *velero.RestoreItemActio
 			deletionPolicy = string(snapshotv1api.VolumeSnapshotContentRetain)
 		}
 
+		if deletionPolicy == string(snapshotv1api.VolumeSnapshotContentDelete) {
```
reasonerjt, line 96: why not just ignore the annotation and set the deletion policy always to retain, and print a explicit message.

reasonerjt:
Could you also update the doc, to point this out in `Implementation choices`:
https://github.com/vmware-tanzu/velero/blob/4cca3996f507a170561f4683f2c47427e7b541e3/site/content/docs/main/csi.md#L36

internal/restore/volumesnapshot_action.go
blackpiglet: Reasonable. My thought is limited the log print numbers.

reasonerjt:
1.  Please also revise the commit msg to explicitly point out that this change is for restore process.
2.  Please double check if the change to handle vsc resource is necessary.

ywk253100 (approved):

PR #115, opened 2022-06-29
support GA storage-provider annotation

luca-heitmann:
Hi, the storage-provider annotation "volume.kubernetes.io/storage-provisioner" is now GA, so it is necessary to delete it on restore as well. I noticed it when I tested to backup/restore Longhorn PVs. This was not possible without this change.

related issue: https://github.com/kubernetes/kubernetes/issues/102357

### DIFF
--- a/internal/restore/pvc_action.go
+++ b/internal/restore/pvc_action.go
@@ -35,10 +35,11 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


innobead:
cc @jenting

jenting (approved):

blackpiglet (approved):

PR #116, opened 2022-08-20
Build and publish multi-arch images

blampe:
This builds the binary as part of the Docker build process instead of copying the binary into the container. I've used this successfully and wanted to share if it would be useful.

Fixes https://github.com/vmware-tanzu/velero/issues/4303.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+Dockerfile
+Makefile
+_output
+velero-plugin-for-csi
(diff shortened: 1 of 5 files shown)


ryanmcafee:
@skriss @carlisia can you guys review this pr? We would like to enable this csi plugin in our AWS EKS clusters, but are running Graviton instances and require linux/arm64 support in the velero-plugin-for-csi container (via docker buildx). We are currently blocked without this addition and would prefer to not have to fork this repo and maintain this limited feature in the GHA workflow. Thanks!

blackpiglet:
I close this PR. The same issue is traced in #151.

PR #118, opened 2022-09-01
Update velero dependency, because BackupItemAction incompatible change.

blackpiglet:
Signed-off-by: Xun Jiang <blackpiglet@gmail.com>


reasonerjt (approved):

sseago:
Lets hold off on this until https://github.com/vmware-tanzu/velero/pull/5263 is merged, which will remove the backwards compatibility problem. We've worked out a way to handle the plugin versioning without breaking old plugins -- basically the v1 proto-generated go files will remain where they are, and only v2 and later will move to new packages. This allows v1 plugins compiled with earlier velero versions to continue to work.

PR #120, opened 2022-09-21
Update go build version?

xchsxhxxbrxhxm:
We use continuous scanning with tools to scan our hardened images. We are encountering a number of CVEs regarding go1.17.11. Will there be a future release updating go version? Let me know if you need more information about the CVEs.


blackpiglet:
Usually, Velero and plugins branch don't pin a specific golang minor version like the release tag, and Velero is going to adopt Golang v1.18.

PR #124, opened 2022-10-13
Remove Restic reference.

blackpiglet:
Remove Restic reference in CSI plugin's code and document.
Resolve some warings in `util_test.go` file

Signed-off-by: Xun Jiang <blackpiglet@gmail.com>


Lyndon-Li (approved):

go.mod
```diff
@@ -8,71 +8,114 @@ require (
 	github.com/sirupsen/logrus v1.8.1
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.7.1
-	github.com/vmware-tanzu/velero v1.8.1
```
shubham-pampattiwar, line 11: @blackpiglet Is this go.mod change intentional ? Should this be pinned to specific version or latest ?

sseago: @shubham-pampattiwar looks like that's needed because it's using the new name `IsPVCDefaultToFSBackup` in the PR. I'm guessing this will need to change to `v1.10.0` once we release 1.10. There's no such thing as "latest" in go.mod after running `go mod tidy` -- you can either reference a tag or a commit, as `tidy` replaces branch references with a specific commit hash, but we won't have a tag until we release.

shubham-pampattiwar: @sseago yeah makes sense, will need a follow up PR to update the go.mod once branches are created.

sseago: @shubham-pampattiwar not branch creation but release tag creation.

blackpiglet: Right. Also needs recently merged RestoreItemAction's change to make the CSI plugin work.
I will modify the reference of Velero package once v1.10 is released.

qiuming-best (approved):

PR #127, opened 2022-11-01
Honour restorePV flag in restore flow

anshulahuja98:
Fixes https://github.com/vmware-tanzu/velero/issues/5527

### DIFF
--- a/internal/restore/pvc_action.go
+++ b/internal/restore/pvc_action.go
@@ -32,6 +32,7 @@ import (
 
 	"github.com/vmware-tanzu/velero-plugin-for-csi/internal/util"
 	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
+	"github.com/vmware-tanzu/velero/pkg/util/boolptr"
 )
 
 const (
@@ -103,6 +104,10 @@ func (p *PVCRestoreItemAction) Execute(input *velero.RestoreItemActionExecuteInp
 		return nil, errors.WithStack(err)
 	}
 	p.Log.Infof("Starting PVCRestoreItemAction for PVC %s/%s", pvc.Namespace, pvc.Name)
+	if boolptr.IsSetToFalse(input.Restore.Spec.RestorePVs) {
+		p.Log.Infof("Restore did not request for PVs to be restored %s/%s", input.Restore.Namespace, input.Restore.Name)
+		return &velero.RestoreItemActionExecuteOutput{SkipRestore: true}, nil
+	}
 
 	removePVCAnnotations(&pvc,
 		[]string{AnnBindCompleted, AnnBoundByController, AnnStorageProvisioner, AnnBetaStorageProvisioner, AnnSelectedNode})
--- a/internal/restore/volumesnapshot_action.go
+++ b/internal/restore/volumesnapshot_action.go
@@ -32,6 +32,7 @@ import (
 	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	"github.com/vmware-tanzu/velero/pkg/label"
 	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
+	"github.com/vmware-tanzu/velero/pkg/util/boolptr"
 )
 
 // VolumeSnapshotRestoreItemAction is a Velero restore item action plugin for VolumeSnapshots
@@ -62,6 +63,10 @@ func resetVolumeSnapshotAnnotation(vs *snapshotv1api.VolumeSnapshot) {
 // to recreate a volumesnapshotcontent object and statically bind the Volumesnapshot object being restored.
 func (p *VolumeSnapshotRestoreItemAction) Execute(input *velero.RestoreItemActionExecuteInput) (*velero.RestoreItemActionExecuteOutput, error) {
 	p.Log.Info("Starting VolumeSnapshotRestoreItemAction")
+	if boolptr.IsSetToFalse(input.Restore.Spec.RestorePVs) {
+		p.Log.Infof("Restore did not request for PVs to be restored %s/%s", input.Restore.Namespace, input.Restore.Name)
+		return &velero.RestoreItemActionExecuteOutput{SkipRestore: true}, nil
+	}
 	var vs snapshotv1api.VolumeSnapshot
 
 	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(input.Item.UnstructuredContent(), &vs); err != nil {
--- a/internal/restore/volumesnapshotclass_action.go
+++ b/internal/restore/volumesnapshotclass_action.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/vmware-tanzu/velero-plugin-for-csi/internal/util"
 	"github.com/vmware-tanzu/velero/pkg/plugin/velero"
+	"github.com/vmware-tanzu/velero/pkg/util/boolptr"
 )
 
 // VolumeSnapshotClassRestoreItemAction is a Velero restore item action plugin for VolumeSnapshotClass
@@ -44,7 +45,10 @@ func (p *VolumeSnapshotClassRestoreItemAction) AppliesTo() (velero.ResourceSelec
 // Execute restores volumesnapshotclass objects returning any snapshotlister secret as additional items to restore
 func (p *VolumeSnapshotClassRestoreItemAction) Execute(input *velero.RestoreItemActionExecuteInput) (*velero.RestoreItemActionExecuteOutput, error) {
 	p.Log.Info("Starting VolumeSnapshotClassRestoreItemAction")
-
+	if boolptr.IsSetToFalse(input.Restore.Spec.RestorePVs) {
+		p.Log.Infof("Restore did not request for PVs to be restored %s/%s", input.Restore.Namespace, input.Restore.Name)
+		return &velero.RestoreItemActionExecuteOutput{SkipRestore: true}, nil
+	}
 	var snapClass snapshotv1api.VolumeSnapshotClass
 
 	if err := runtime.DefaultUnstructuredConverter.FromUnstructured(input.Item.UnstructuredContent(), &snapClass); err != nil {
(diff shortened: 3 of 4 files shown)


internal/restore/pvc_action.go
```diff
@@ -103,6 +104,10 @@ func (p *PVCRestoreItemAction) Execute(input *velero.RestoreItemActionExecuteInp
 		return nil, errors.WithStack(err)
 	}
 	p.Log.Infof("Starting PVCRestoreItemAction for PVC %s/%s", pvc.Namespace, pvc.Name)
+	if boolptr.IsSetToFalse(input.Restore.Spec.RestorePVs) {
+		p.Log.Infof("Restore did not request for PVs to be restored %s/%s", input.Restore.Namespace, input.Restore.Name)
+		return &velero.RestoreItemActionExecuteOutput{SkipRestore: true}, nil
```
sshende-catalogicsoftware, line 109: Do we want to skip the entire PVC here?

How about we allow the PVC to be restored, but without the change in datasource? We can skip resource for the Volumesnapshot and VolumesnapshotContent resource, but we should let the PVC be provisioned. It will have an empty volume which should be the intended behavior?

Actually, the flag "input.Restore.Spec.RestorePVs" should mean "input.Restore.Spec.RestorePVsFrom Snapshots" as we want this flag to only control if the data from snapshots needs to be restored, not the Persistent Volumes themselves?

Suggestion:
`return &velero.RestoreItemActionExecuteOutput{
			UpdatedItem: input.Item,
		}, nil`

anshulahuja98: I think that makes sense. 
I will make this change as you suggested.

sshende-catalogicsoftware: Ok, if this is the approach then I think you should not be skipping restore of volumesnapshot and volumesnapshotcontent as well. 

Consider a case where the backup is from a restored PVC(CSI). The PVC will have a volumesnapshot as the datasource. Now, if this approach is implemented, the PVC will still have the old datasource pointing to some volumesnapshot, but, we are skipping the restoration of volumesnapshots. In this case, the restore will succeed but the PVC would never come up. Consider the case where the restore is ahppening to an alternate namespace or to a namespace that does not already have the volumesnapshot resource in the namespace. 

So, In my opinion, if we take this approach, we should remove the code to skip restore for Volumesnapshots and VolumeSnapshotcontents as well.

What do you think?

anshulahuja98: "The PVC will have a volumesnapshot as the datasource. "

Why will that happen?

We will pretty much provision a clean PVC right, without any datasource. 
Since we are just returning the input.item as UpdateItem.

Basically, making it a no op.

sshende-catalogicsoftware: This can happen if you are trying to backup  a PVC which has been restored from a CSI snapshot earlier.

anshulahuja98: Ah I see.
What if we add some code to handle this scenario

If restore PV = false, in the input item, we'll remove the Source.

Does that help?

shubham-pampattiwar:
@anshulahuja98 I am unsure about these changes, `restorePVs` from the restore API is used to specify whether to restore all the PVs from snapshot, this snapshot is a cloudprovider snapshot and not a CSI snapshot. AFAIK `restorePVs` flag is not related to CSI Snapshots.

anshulahuja98:
> @anshulahuja98 I am unsure about these changes, `restorePVs` from the restore API is used to specify whether to restore all the PVs from snapshot, this snapshot is a cloudprovider snapshot and not a CSI snapshot. AFAIK `restorePVs` flag is not related to CSI Snapshots.

@shubham-pampattiwar , this is inline with the snapshotVolumes flag in the Backup API: https://github.com/vmware-tanzu/velero-plugin-for-csi/blob/3c700d51714c637947f6f93b6b642267bbaa4327/internal/backup/pvc_action.go#L61

Hence naturally restorePVs flag should also be honored.

Otherwise, there is no way for a user to restore Velero backup without Volume data. Very hard to take care of basic dev/test replication scenario.

shubham-pampattiwar:
@anshulahuja98 I believe you can use `excludedResources:` to specify which resource(s) you do not want to restore? Would that not satisfy your usecase ?

anshulahuja98:
> @anshulahuja98 I believe you can use `excludedResources:` to specify which resource(s) you do not want to restore? Would that not satisfy your usecase ?

I just feel for a more consistent experience this change should be made. 
If snapshotVolumes is honored then restorePV should be honored, otherwise there is no parity.

Even from CLI POV, restore PV false behavior should be deterministic, and not random

CSI is anyways going to be the main volume driver going forward, hence consistency of experience should accrue to it.

Can you please explain why this change does not seem appropriate to you?

Also the other scenario that @sshende-catalogicsoftware  pointed out above, we need scenarios where only PVCs can be restored.  This can be achieved using this flag. Where we patch the PVC to remove the source, and skip restoring VS, VSC and PV.

That can't be achieved properly through excludeResources filter. 
Since the PVC might be an already restored PVC with source information. And hence will trigger data copy if PV is excluded. As part of this PR, we can write code to remove the snapshot source.

anshulahuja98:
CC: @blackpiglet

anshulahuja98:
CC: @sseago

blackpiglet:
I'm also not sure about this change. May need to take more consideration on backup's related flag `SnapshotVolumes`.
```
	// SnapshotVolumes specifies whether to take cloud snapshots
	// of any PV's referenced in the set of objects included
	// in the Backup.
	// +optional
	// +nullable
	SnapshotVolumes *bool `json:"snapshotVolumes,omitempty"`
```
`SnapshotVolumes` is used to indicate whether to take snapshot with cloud provider plugin, not for CSI plugin, and `RestorePVs` works with it.

```
	// RestorePVs specifies whether to restore all included
	// PVs from snapshot (via the cloudprovider).
	// +optional
	// +nullable
	RestorePVs *bool `json:"restorePVs,omitempty"`
``` 

I am not sure about whether needs a new flag to disable taking and resoting all CSI snapshots, even needs a new one, I think maybe adding a new flag is better.

anshulahuja98:
> ```
> 	SnapshotVolumes *bool `json:"snapshotVolumes,omitempty"`
> ```
> 
> 
>     
>   
> 
> `SnapshotVolumes` is used to indicate whether to take snapshot with cloud provider plugin, not for CSI plugin, and `RestorePVs` works with it.

"	SnapshotVolumes *bool `json:"snapshotVolumes,omitempty"`
SnapshotVolumes is used to indicate whether to take snapshot with cloud provider plugin, not for CSI plugin, and RestorePVs works with it."

@blackpiglet : snapshot volumes is already honored by CSI Plugin

As per the below code check.
https://github.com/vmware-tanzu/velero-plugin-for-csi/blob/3c700d51714c637947f6f93b6b642267bbaa4327/internal/backup/pvc_action.go#L61

So inline with it restorePV should also be honored right

If snapshotVolumes was not honored by CSI Plugin then we could have debated further for both flags.

blackpiglet:
Found this part.

```
	// Do nothing if volume snapshots have not been requested in this backup
	if boolptr.IsSetToFalse(backup.Spec.SnapshotVolumes) {
		p.Log.Infof("Volume snapshotting not requested for backup %s/%s", backup.Namespace, backup.Name)
		return item, nil, nil
	}
```

@shubham-pampattiwar What's your opinion about this change?

shubham-pampattiwar:
`SnapshotVolumes` in backup spec - SnapshotVolumes specifies whether to take **cloud provider based snapshots** of any PV's referenced in the set of objects included in the Backup. This flag does not trigger **CSI volume snapshots**.

`RestorePVs` in restore spec - RestorePVs specifies whether to restore all included PVs from **snapshot (via the cloudprovider).** This flag **does not** trigger the restore of PVs from backed up **CSI snapshots**

IMHO we should keep these flags as is and not use them in the context of CSI volumes B/R. I like the idea of using a separate flag for CSI Volumes.

anshulahuja98:
@shubham-pampattiwar 
"SnapshotVolumes in backup spec - SnapshotVolumes specifies whether to take cloud provider based snapshots of any PV's referenced in the set of objects included in the Backup. This flag does not trigger CSI volume snapshots."

As per the snippet @blackpiglet shared above, it does trigger CSI snapshot.  Or is my understanding of that snippet different?

Or do you propose removing it from CSI Plugin?

IMO Introducing new flags for each plugin might bloat up the CR contract. And since everyone might not use a particular plugin, it might lead to confusion.

shubham-pampattiwar:
I agree with the fact the we should not bloat up the backup CR with new flags but IMHO we should not keep cloud provider based snapshots as well as  CSI snapshots tied up with the same flag.

sseago:
@shubham-pampattiwar If we decide to use separate flags for CSI and snapshots restore, we may want to disentangle SnapshotVolumes on the backup side as well, which currently applies to both snapshot and CSI. So if we add a new flag for CSI, the same change should probably add a new flag on the backup side for CSI and stop honoring SnapshotVolumes. Also, since CSI is a plugin and not in core velero, any "flag" for CSI probably needs to be an annotation on the Backup and Restore CRs, not a spec flag -- I think it would be bad precedent to introduce a CRD change that is only relevant to a single plugin.

anshulahuja98:
Hi @sseago / @shubham-pampattiwar / @blackpiglet  / @OrlinVasilev 
Any closure on this PR?
Been tracking this since 3 months, any closure would be helpful :)

shubham-pampattiwar:
@anshulahuja98 Thank you for the PR, as discussed in the community call let go ahead with this approach.

shubham-pampattiwar (approved):

shubham-pampattiwar (changes requested):
@anshulahuja98 Can you please sign the commit and update the PR ?

anshulahuja98:
Request to pl re approve @shubham-pampattiwar, completed DCO
REquest additional 2nd approval from @blackpiglet / @sseago

shubham-pampattiwar (approved):

sseago (approved):

PR #154, opened 2023-04-03
When restorePVs is false, CSI should restore the PVC.

blackpiglet:
The PVC's data source and referenced volume should all be cleaned for restore to let k8s to create a new PV and volume.


shubham-pampattiwar (approved):
Thanks @blackpiglet , LGTM!

PR #177, opened 2023-05-25
Make the BIA in CSI plugin adopt v2.

blackpiglet:
Move the polling VS ReadyToUse logic into Async operation.


internal/backup/volumesnapshot_action.go
```diff
@@ -149,13 +153,78 @@ func (p *VolumeSnapshotBackupItemAction) Execute(item runtime.Unstructured, back
...
 
 	p.Log.Infof("Returning from VolumeSnapshotBackupItemAction with %d additionalItems to backup", len(additionalItems))
 	for _, ai := range additionalItems {
 		p.Log.Debugf("%s: %s", ai.GroupResource.String(), ai.Name)
 	}
 
-	return &unstructured.Unstructured{Object: vsMap}, additionalItems, nil
+	// The operationID is of the form <namespace>/<volumesnapshot-name>
+	operationID := vs.Namespace + "/" + vs.Name
+	itemToUpdate := []velero.ResourceIdentifier{
```
sseago, line 166: Since you're returning the current resource (the VolumeSnapshot) to update during finalize, this plugin will be called again, so at the beginning of the Execute func, you need to check backup.Phase and if it's Finalizing or FinalizingPartiallyFailed, then just return the input item without any changes or actions since all of the Execute actions already happened during InProgress.

internal/backup/volumesnapshot_action.go
```diff
@@ -149,13 +153,78 @@ func (p *VolumeSnapshotBackupItemAction) Execute(item runtime.Unstructured, back
...
+	if vs.Status.CreationTime != nil {
+		progress.Started = vs.Status.CreationTime.Time
+	}
+
+	if boolptr.IsSetToTrue(vs.Status.ReadyToUse) {
+		progress.Completed = true
+	} else if vs.Status.Error != nil {
+		progress.Completed = true
+		if vs.Status.Error.Time != nil {
+			progress.Updated = vs.Status.Error.Time.Time
```
sseago, line 217: It might be clearer if progress.Updated is always set to current time rather than to error time (and leaving it blank if no error). Generally, the intent of updated is so that users can know when the last update occurred (and, for completed operations, when the completion field was set, since we won't call Progress again once completed returns as true).

blackpiglet: Thanks for pointing this out. Removed the Updated time setting.

blackpiglet: Also, remove the start time, because VolumeSnapshot has no start time for the uploading phase;
VolumeSnapshot creation time is not accurate for the operation. It can be earlier than the operation creation time.

internal/backup/volumesnapshot_action.go
blackpiglet: Thanks. Added the checking logic.

blackpiglet:
@sseago 
I met a problem when removing the polling logic in the Velero server code.
In the Velero server backup controller code, there was a polling logic of VolumeSnapshot status turning into `ReadyToUse`.
 And there is also logic to delete VolumeSnapshot after the VolumeSnapshot turned into `ReadyToUse`.
https://github.com/vmware-tanzu/velero/blob/main/pkg/controller/backup_controller.go#L994-L1000 

 I'm not sure about whether to move the deleing VolumeSnapshot logic to the BIA v2 also.
Because the deleting logic has max concurrent connections limitation with kube-apiserver, it looks like there is no way to limit the concurrent BIA v2 actions. It's possible deleting VolumeSnapshot logic in BIA v2 may overload the kube-apiserver with a big number of VolumeSnapshots in backups.

What's your opinion about this? Is it possible to limit the concurrent BIA v2 action numbers?
If not, where should we put the deleting VolumeSnapshots logic? I think the backup_operations_controller.go is an option.

@reasonerjt @Lyndon-Li

internal/backup/volumesnapshotcontent_action.go
```diff
@@ -73,9 +74,26 @@ func (p *VolumeSnapshotContentBackupItemAction) Execute(item runtime.Unstructure
 
 	snapContMap, err := runtime.DefaultUnstructuredConverter.ToUnstructured(&snapCont)
 	if err != nil {
-		return nil, nil, errors.WithStack(err)
+		return nil, nil, "", nil, errors.WithStack(err)
 	}
 
 	p.Log.Infof("Returning from VolumeSnapshotContentBackupItemAction with %d additionalItems to backup", len(additionalItems))
-	return &unstructured.Unstructured{Object: snapContMap}, additionalItems, nil
+	return &unstructured.Unstructured{Object: snapContMap}, additionalItems, "", nil, nil
```
Lyndon-Li, line 81: Do we need to add the VSC object to itemToUpdate as well? This is decided by two factors:
1. Whether the VSC object is modified during the async operation -- wait for it becomes readyToUse?  (I think the answer is yes)
2. Whether the changes in 1 is critical to Velero so they must be persisted

blackpiglet: https://github.com/kubernetes-csi/external-snapshotter/blob/master/client/apis/volumesnapshot/v1/types.go#L121-L130
Per the external-snapshotter comments, the `ReadyToUse` in VolumeSnapshotStatus and VolumeSnapshotContentStauts should be the same, and fields in VolumeSnapshotStatus are updated based on fields in VolumeSnapshotContentStatus.
I think just watching over VolumeSnapshotContentStauts's change should be enough.

Lyndon-Li: Make one item as itemToUpdate means the item will be update to backup tarball the second time after all async operation finishes.
Or in another words, if we don't need to save the up-to-date object in backup tarball, we don't even need to make VS object as an itemToUpdate either.

internal/backup/volumesnapshot_action.go
sseago: If we don't have a real start time, I wonder whether it makes sense to use the current time as the start time (on first progress call)? I just wonder whether an empty start time might break something -- in any case we should make sure that all of the velero core logic doesn't break if start time is empty, since there's no guarantee that a plugin would fill it in.

internal/backup/volumesnapshotcontent_action.go
sseago: @Lyndon-Li The volumesnapshot will be in the tarball in any case because the PVC plugin returns it as an additional item. It is the *current item* handled by this plugin, which means that after this plugin execute exits, Velero adds this item to the backuup. Adding it to itemsToUpdate just means that we get an updated copy of this resource during finalize. If we don't add it to itemsToUpdate, then the tarball will have the not-yet-ready version.

@shubham-pampattiwar the current OADP datamover code requires the VS and VSC to be in the tarball on restore, right? Let me know if I'm misremembering.

sseago:
@blackpiglet backup operations controller isn't a good place for it, since that controller won't act on a backup if all async operations are completed when leaving the InProgress phase -- if all async operations are done when the backup controller reconcile is complete (i.e. upon uploading tarball, logs, etc.), then we go straight to finalize. Finalize is never skipped, so that's the place we want this. We had to do exactly this in our OADP velero fork to deal with the OADP datamover, since we needed VS to not be deleted during datamover plugin operations:
https://github.com/vmware-tanzu/velero/commit/8e52bcaf9b9ef0f35e83624e82bac7392f615c38

internal/backup/volumesnapshotcontent_action.go
Lyndon-Li: @sseago 
Exactly, by that I meant, we need to put the VSC object into itemToUpdate as same as VS object

Lyndon-Li:
@sseago 
We can delete the VS in finalize phase, e.g. inside `FinalizeBackup` function.
On the other hand, since the VS is in itemToUpdate list, the VS object will be called against BIA Execute for the second time, this is also an opportunity.

Ideally, we take the second way, since the snapshot is created by the plugin so it make more sense to delete it by plugin as well.
Let us know if that works with OADP data mover or not.

internal/backup/volumesnapshot_action.go
```diff
@@ -54,17 +57,26 @@ func (p *VolumeSnapshotBackupItemAction) AppliesTo() (velero.ResourceSelector, e
...
+func (p *VolumeSnapshotBackupItemAction) Execute(item runtime.Unstructured, backup *velerov1api.Backup) (runtime.Unstructured, []velero.ResourceIdentifier,
+	string, []velero.ResourceIdentifier, error) {
 	p.Log.Infof("Executing VolumeSnapshotBackupItemAction")
 
+	if backup.Status.Phase == velerov1api.BackupPhaseFinalizing ||
+		backup.Status.Phase == velerov1api.BackupPhaseFinalizingPartiallyFailed {
+		p.Log.WithField("Backup", fmt.Sprintf("%s/%s", backup.Namespace, backup.Name)).
+			WithField("BackupPhase", backup.Status.Phase).
+			Info("Backup is in finalizing phase. Skip the Action.")
+		return item, nil, "", nil, nil
```
sseago, line 69: @Lyndon-Li I see what you mean now about VS deletion during finalize. The deletion code would go right here before the return line.

sseago: @shubham-pampattiwar This should be fine for OADP too, right? It puts VS deletion in finalize, same phase as in our fork, and with this we could drop the fork commit.

sseago: And yes, this is cleaner than the current (1.11) way of deleting something in core backup code that was created in a plugin.

internal/backup/volumesnapshotcontent_action.go
blackpiglet: I see. I will add the VSC in the itemsToUpdate to make sure its state is updated after the operation completes.

internal/backup/volumesnapshot_action.go
blackpiglet: In the CSI case, the `Progress()` function cannot tell whether it is running the first time.

I can only find one place to store the start time. The `operationID` parameter in `Progress()` function.
Setting it to the format `operationID := vs.Namespace + "/" + vs.Name + "/" + startedTime` can make it work.

internal/backup/volumesnapshotcontent_action.go
```diff
@@ -47,12 +55,18 @@ func (p *VolumeSnapshotContentBackupItemAction) AppliesTo() (velero.ResourceSele
 
 // Execute returns the unmodified volumesnapshotcontent object along with the snapshot deletion secret, if any, from its annotation
 // as additional items to backup.
-func (p *VolumeSnapshotContentBackupItemAction) Execute(item runtime.Unstructured, backup *velerov1api.Backup) (runtime.Unstructured, []velero.ResourceIdentifier, error) {
+func (p *VolumeSnapshotContentBackupItemAction) Execute(item runtime.Unstructured, backup *velerov1api.Backup) (runtime.Unstructured, []velero.ResourceIdentifier, string, []velero.ResourceIdentifier, error) {
 	p.Log.Infof("Executing VolumeSnapshotContentBackupItemAction")
 
+	if backup.Status.Phase == velerov1api.BackupPhaseFinalizing || backup.Status.Phase == velerov1api.BackupPhaseFinalizingPartiallyFailed {
+		p.Log.WithField("Backup", fmt.Sprintf("%s/%s", backup.Namespace, backup.Name)).
```
Lyndon-Li, line 62: Change it to a Debug log? It looks like a trivial log

Lyndon-Li:
Do we have a Velero counterpart for this PR? Since we have moved some code from Velero

internal/backup/volumesnapshotcontent_action.go
blackpiglet: Modified.

internal/backup/volumesnapshot_action.go
```diff
@@ -144,18 +154,98 @@ func (p *VolumeSnapshotBackupItemAction) Execute(item runtime.Unstructured, back
...
+	}
+
+	vs, err := snapshotClient.SnapshotV1().VolumeSnapshots(operationIDParts[0]).Get(context.Background(), operationIDParts[1], metav1.GetOptions{})
+	if err != nil {
+		p.Log.Errorf("error getting volumesnapshot %s/%s: %s", operationIDParts[0], operationIDParts[1], err.Error())
+		return progress, errors.WithStack(err)
+	}
+
+	if vs.Status == nil {
+		return progress, nil
```
shubham-pampattiwar, line 233: maybe add a debug log here that VS is yet to have a status or something

internal/backup/volumesnapshot_action.go
```diff
@@ -93,13 +98,18 @@ func (p *VolumeSnapshotBackupItemAction) Execute(item runtime.Unstructured, back
...
 	if err != nil {
 		util.CleanupVolumeSnapshot(&vs, snapshotClient.SnapshotV1(), p.Log)
-		return nil, nil, errors.WithStack(err)
+		return nil, nil, "", nil, errors.WithStack(err)
 	}
 
-	annotations := map[string]string{
-		util.MustIncludeAdditionalItemAnnotation: "true",
+	if backup.Status.Phase == velerov1api.BackupPhaseFinalizing || backup.Status.Phase == velerov1api.BackupPhaseFinalizingPartiallyFailed {
+		p.Log.WithField("Backup", fmt.Sprintf("%s/%s", backup.Namespace, backup.Name)).
```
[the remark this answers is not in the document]
sseago: Actually, instead of removing it, if you decide you don't want it to log every time, make it a debug log.

internal/backup/volumesnapshotcontent_action.go
```diff
@@ -73,9 +87,72 @@ func (p *VolumeSnapshotContentBackupItemAction) Execute(item runtime.Unstructure
...
+	}
+
+	vsc, err := snapshotClient.SnapshotV1().VolumeSnapshotContents().Get(context.Background(), operationsIDParts[0], metav1.GetOptions{})
+	if err != nil {
+		p.Log.Errorf("error getting volumesnapshotcontent %s: %s", operationsIDParts[0], err.Error())
+		return progress, errors.WithStack(err)
+	}
+
+	if vsc.Status == nil {
+		return progress, nil
```
shubham-pampattiwar, line 140: maybe add a debug log here that VSC is yet to have a status or something

blackpiglet: Done

internal/backup/volumesnapshot_action.go
blackpiglet: Done

internal/backup/volumesnapshot_action.go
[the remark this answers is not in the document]
blackpiglet: Changed to debug log.

shubham-pampattiwar:
@blackpiglet @Lyndon-Li Will there be a separate PR for datamover changes relaed to CSI plugin ?

blackpiglet:
> @blackpiglet @Lyndon-Li Will there be a separate PR for datamover changes relaed to CSI plugin ?

@shubham-pampattiwar Sure. I'm working on it. The BIA part is also in progress, and I will post a PR today, and I think most of rest work should be done in this week.

Lyndon-Li (approved):

internal/backup/volumesnapshot_action.go
```diff
@@ -93,13 +98,18 @@ func (p *VolumeSnapshotBackupItemAction) Execute(item runtime.Unstructured, back
 	vsc, err := util.GetVolumeSnapshotContentForVolumeSnapshot(&vs, snapshotClient.SnapshotV1(), p.Log, backupOngoing)
 	if err != nil {
 		util.CleanupVolumeSnapshot(&vs, snapshotClient.SnapshotV1(), p.Log)
-		return nil, nil, errors.WithStack(err)
+		return nil, nil, "", nil, errors.WithStack(err)
 	}
 
-	annotations := map[string]string{
-		util.MustIncludeAdditionalItemAnnotation: "true",
+	if backup.Status.Phase == velerov1api.BackupPhaseFinalizing || backup.Status.Phase == velerov1api.BackupPhaseFinalizingPartiallyFailed {
```
shubham-pampattiwar, line 104: +1 @blackpiglet

shubham-pampattiwar (approved):

PR #181, opened 2023-06-06
Support DataUpload CR handling in backup

blackpiglet:
Fix https://github.com/vmware-tanzu/velero/issues/6114


internal/backup/pvc_action.go
```diff
@@ -181,9 +235,148 @@ func (p *PVCBackupItemAction) Progress(operationID string, backup *velerov1api.B
 		return progress, biav2.InvalidOperationIDError(operationID)
 	}
 
+	dataUpload, err := getDataUpload(context.Background(), backup, p.VeleroClient, operationID)
+	if err != nil {
+		p.Log.Errorf("fail to get DataUpload for backup %s/%s: %s", backup.Namespace, backup.Name, err.Error())
+		return progress, err
+	}
+
```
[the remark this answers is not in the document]
blackpiglet: Sure. Will add the checks.

internal/util/util.go
```diff
@@ -433,3 +445,18 @@ func recreateVolumeSnapshotContent(vsc snapshotv1api.VolumeSnapshotContent, back
 
 	return nil
 }
+
+func DeleteVolumeSnapshotIfAny(ctx context.Context, snapshotClient snapshotterClientSet.Interface,
+	vs snapshotv1api.VolumeSnapshot, log logrus.FieldLogger) {
+	_, err := snapshotClient.SnapshotV1().VolumeSnapshots(vs.Namespace).Get(ctx, vs.Name, metav1.GetOptions{})
```
Lyndon-Li, line 451: This can be simplified --- directly call `snapshotClient.SnapshotV1().VolumeSnapshots(vs.Namespace).Delete` and check for `apierrors.IsNotFound(err)` and log messages accordingly.

internal/util/util.go
```diff
@@ -433,3 +445,18 @@ func recreateVolumeSnapshotContent(vsc snapshotv1api.VolumeSnapshotContent, back
...
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			log.WithError(err).Errorf("fail to get VolumeSnapshot %s/%s", vs.Namespace, vs.Name)
```
Lyndon-Li, line 454: This is not an error log necessarily since this function is called by "IfAny", so suggest to change its level, e.g., debug level.
Otherwise, since the VS must exist in the context this function is called, we just modify the function to "
DeleteVolumeSnapshot" and log error message on any errors? then we don't need to check `apierrors.IsNotFound(err)`

blackpiglet: This function is called to clear related VolumeSnapshot when DataUpload creation failed.
Whether it fails or not, the BIA is returning with the previous error.
I think setting the log level to info or warn should be enough in this case.
Users can see the information if VolumeSnapshot is not cleaned.

Lyndon-Li (approved):

qiuming-best (approved):

blackpiglet:
Also fix https://github.com/vmware-tanzu/velero/issues/6116

PR #184, opened 2023-06-15
Add snapshot data mover logic in PVC RIA v2.

blackpiglet:
Fix https://github.com/vmware-tanzu/velero/issues/6115

### DIFF
--- a/internal/restore/pvc_action.go
+++ b/internal/restore/pvc_action.go
@@ -21,16 +21,18 @@ import (
 	"encoding/json"
 	"fmt"
 
+	jsonpatch "github.com/evanphx/json-patch"
 	snapshotv1api "github.com/kubernetes-csi/external-snapshotter/client/v4/apis/volumesnapshot/v1"
 	snapshotterClientSet "github.com/kubernetes-csi/external-snapshotter/client/v4/clientset/versioned"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-
 	corev1api "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	utilrand "k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/client-go/kubernetes"
 
 	"github.com/vmware-tanzu/velero-plugin-for-csi/internal/util"
@@ -49,6 +51,10 @@ const (
 	AnnStorageProvisioner     = "volume.kubernetes.io/storage-provisioner"
 	AnnBetaStorageProvisioner = "volume.beta.kubernetes.io/storage-provisioner"
 	AnnSelectedNode           = "volume.kubernetes.io/selected-node"
+)
+
+const (
+	GenerateNameRandomLength = 5
 )
 
 // PVCRestoreItemAction is a restore item action plugin for Velero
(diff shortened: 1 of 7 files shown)


internal/restore/pvc_action.go
```diff
@@ -189,19 +191,81 @@ func (p *PVCRestoreItemAction) Progress(operationID string, restore *velerov1api
...
+	return true, nil
 }
 
 func getDataUploadResult(ctx context.Context, restore *velerov1api.Restore, pvc *corev1api.PersistentVolumeClaim,
 	kubeClient kubernetes.Interface) (*velerov2alpha1.DataUploadResult, error) {
 	cmList, err := kubeClient.CoreV1().ConfigMaps(restore.Namespace).List(ctx, metav1.ListOptions{
-		LabelSelector: fmt.Sprintf("%s=%s,%s=%s", velerov1api.BackupNameLabel, label.GetValidName(restore.Spec.BackupName),
-			util.PVCNamespaceNameLabel, pvc.Namespace+"/"+pvc.Name),
+		LabelSelector: fmt.Sprintf("%s=%s,%s=%s", util.PVCNamespaceNameLabel, pvc.Namespace+"."+pvc.Name,
+			velerov1api.BackupNameLabel, label.GetValidName(restore.Spec.BackupName)),
```
Lyndon-Li, line 268: As we discussed, this label cannot uniquely identify a cm, since two restores of the same PVC from the same backup will get the same label. So probably, we need to add the restore name to the label?

internal/restore/pvc_action.go
```diff
@@ -228,3 +292,145 @@ func getDataUploadResult(ctx context.Context, restore *velerov1api.Restore, pvc
...
+		return nil, errors.Wrapf(err, "fail get DataUploadResult for restore: %s", restore.Name)
+	}
+	pvc.Spec.VolumeName = ""
+	if pvc.Spec.Selector == nil {
+		pvc.Spec.Selector = &metav1.LabelSelector{}
+	}
+	if pvc.Spec.Selector.MatchLabels == nil {
+		pvc.Spec.Selector.MatchLabels = make(map[string]string)
+	}
+	pvc.Spec.Selector.MatchLabels[util.DynamicPVRestoreLabel] = label.GetValidName(fmt.Sprintf("%s.%s", pvc.Namespace, pvc.Name))
```
Lyndon-Li, line 427: We may also need to enhance here, think below scenarios:
One restore is successfully created from the backup, so there is a PV with label `<pvc.Namespace>.<pvc.Name>`
Users delete the restored namespace
For some reason, the PV is not successfully deleted
Now do a new restore to the same namesapce, a new PVC is created to be bound by matching the same label `<pvc.Namespace>.<pvc.Name>`
Since there is already a PV with label `<pvc.Namespace>.<pvc.Name>`, the PVC is bound immediately

internal/restore/pvc_action.go
blackpiglet: Restore UID looks better. Is the backup name label used in other places?
How about changing like this?
```
LabelSelector: fmt.Sprintf("%s=%s,%s=%s", util.PVCNamespaceNameLabel, pvc.Namespace+"."+pvc.Name,
			velerov1api.RestoreUIDLabel, label.GetValidName(restore.UID))
```

Lyndon-Li: OK, so the creation counterpart on Velero side should also be changed.

internal/restore/pvc_action.go
blackpiglet: Then we may also need to add the Restore UID here.

Lyndon-Li: Will that be too long for a PV name? Can we remove the pvc.Namespace since for a particular restore, the target namespace should always be the same?

Lyndon-Li: That should work, but we need to take care of two labels all the time. The label `DynamicPVRestoreLabel` is well known since it has been mentioned in the design, so we will take care of it during other backup types, e.g., during PVB, the label will be removed. However, `RestoreUIDLabel` is not mentioned in the design and it is not a good practice to have many labels for the same purpose from the perspective of the design.

Therefore, do we have any other way to make it?

blackpiglet: After discussion, still using the label `DynamicPVRestoreLabel`. The format is changed to `<pvc-namespace>.<pvc-name>.<random-string(length is 5)>`

Lyndon-Li (approved):

internal/restore/pvc_action.go
```diff
@@ -232,3 +296,145 @@ func getDataUploadResult(ctx context.Context, restore *velerov1api.Restore, pvc
...
+			return errors.Wrapf(err, fmt.Sprintf("Failed to parse %s from annotation on Volumesnapshot %s/%s into restore size",
+				vs.Annotations[util.VolumeSnapshotRestoreSize], vs.Namespace, vs.Name))
+		}
+		// It is possible that the volume provider allocated a larger capacity volume than what was requested in the backed up PVC.
+		// In this scenario the volumesnapshot of the PVC will end being larger than its requested storage size.
+		// Such a PVC, on restore as-is, will be stuck attempting to use a Volumesnapshot as a data source for a PVC that
+		// is not large enough.
+		// To counter that, here we set the storage request on the PVC to the larger of the PVC's storage request and the size of the
+		// VolumeSnapshot
+		setPVCStorageResourceRequest(pvc, restoreSize, logger)
```
shubham-pampattiwar, line 410: +1 @blackpiglet

shubham-pampattiwar (approved):

Lyndon-Li (approved):

blackpiglet:
Also fix https://github.com/vmware-tanzu/velero/issues/6116

PR #189, opened 2023-07-07
Return PVC intact when CSI is not present in PVC spec

blackpiglet:
Return PVC intact when CSI is not present in PVC spec to avoid deleting useful annotations: when backing up and restoring with FileSystem, and installing the Velero CSI plugin alongside, changing the PVC's node-selector by ConfigMap can trigger this issue.

### DIFF
--- a/internal/restore/pvc_action.go
+++ b/internal/restore/pvc_action.go
@@ -167,7 +167,15 @@ func (p *PVCRestoreItemAction) Execute(input *velero.RestoreItemActionExecuteInp
 			}
 			logger.Infof("DataDownload %s/%s is created successfully.", dataDownload.Namespace, dataDownload.Name)
 		} else {
-			if err := restoreFromVolumeSnapshot(&pvc, p.SnapshotClient, logger); err != nil {
+			volumeSnapshotName, ok := pvc.Annotations[util.VolumeSnapshotLabel]
+			if !ok {
+				logger.Info("Skipping PVCRestoreItemAction for PVC , PVC does not have a CSI volumesnapshot.")
+				// Make no change in the input PVC.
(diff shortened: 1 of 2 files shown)


blackpiglet:
change-node-selector action still cannot work properly after this change.
Looks like this is due to the temporary PVC created by the snapshot data mover. Is it possible the change-node-selector restore action is applied to the temporary PVC, not the final PVC?
@Lyndon-Li @danfengliu

internal/restore/pvc_action.go
```diff
@@ -96,6 +96,9 @@ func resetPVCSpec(pvc *corev1api.PersistentVolumeClaim, vsName string) {
 	}
 	pvc.Spec.DataSource = dataSourceRef
 	pvc.Spec.DataSourceRef = dataSourceRef
+
+	// Clean the selector to make the PVC to dynamically allocate PV.
+	pvc.Spec.Selector = nil
```
Lyndon-Li, line 101: Let's avoid do this in a specific RIA, reason:
During restore, Velero calls RIAs recursively. We cannot guarantee that this `pvc.Spec.Selector` is not added by some former RIA. As a result, if we remove it, the restore may encounter problem.

Let's do this in Velero backup instead at here https://github.com/vmware-tanzu/velero/blob/0945879a8a64cfda0fe55875d6451c1f16cc1385/pkg/backup/backup_pv_action.go#L76C1-L76C1, this would solve the similar problems for all backup types in one shot

blackpiglet: Got it. Will create a new PR in Velero repository to address this.

Lyndon-Li (approved):

danfengliu (approved):

PR #193, opened 2023-07-28
Change VolumeSnapshotClassDriverBackupAnnotationPrefix's separator fr…

blackpiglet:
…om "/" to "_"

1. During k8s annotation key format limitation, there is only one "/" allowed. https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set
2. Update referenced Velero tag in go.mod

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.8.2
 	// TODO: need to use velero v1.12.0 after releasing.
-	github.com/vmware-tanzu/velero v0.0.0-20230630063030-84eca51d22c7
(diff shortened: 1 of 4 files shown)


ywk253100 (approved):

blackpiglet:
@anshulahuja98 
I modified the VolumeSnapshotClassDriverBackupAnnotation prefix value for backup because the k8s annotation's key cannot have multiple "/": https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set

I will update the design document accordingly.

anshulahuja98:
Oh good catch. Didn't test this particular one directly in the cluster. 
Thanks for catching this.

Lyndon-Li (approved):

PR #207, opened 2023-10-03
Fix Error log bug

Shashank1306s:
Improved error floatation during csi-polling timeout scenario.

### DIFF
--- a/internal/util/util.go
+++ b/internal/util/util.go
@@ -267,7 +267,7 @@ func GetVolumeSnapshotContentForVolumeSnapshot(volSnap *snapshotv1api.VolumeSnap
 	if err != nil {
 		if err == wait.ErrWaitTimeout {
 			if snapshotContent.Status != nil && snapshotContent.Status.Error != nil {
-				log.Errorf("Timed out awaiting reconciliation of volumesnapshot, Volumesnapshotcontent %s has error: %v", snapshotContent.Name, snapshotContent.Status.Error.Message)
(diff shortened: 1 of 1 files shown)


anshulahuja98 (dismissed):

internal/util/util.go
```diff
@@ -267,7 +267,8 @@ func GetVolumeSnapshotContentForVolumeSnapshot(volSnap *snapshotv1api.VolumeSnap
 	if err != nil {
 		if err == wait.ErrWaitTimeout {
 			if snapshotContent.Status != nil && snapshotContent.Status.Error != nil {
-				log.Errorf("Timed out awaiting reconciliation of volumesnapshot, Volumesnapshotcontent %s has error: %v", snapshotContent.Name, snapshotContent.Status.Error.Message)
+				log.Errorf("Timed out awaiting reconciliation of volumesnapshot, Volumesnapshotcontent %s has error: %v", snapshotContent.Name, *snapshotContent.Status.Error.Message)
+				return nil, errors.Errorf("CSI got timed out with error: %v", *snapshotContent.Status.Error.Message)
```
shubham-pampattiwar, line 271: unsure about this return statement here, we return on line 276 with the original error, right ?

Shashank1306s: Here in line 268 the err is compared to wait.ErrWaitTimeout which is `timed out waiting for the condition`

and while returning this error message is floated in the velero rather than actual error.
So, I've added this return statement for floating actual error

anshulahuja98 (approved):

sseago (approved):

PR #226, opened 2024-01-25
Fix vsc backup plugin progress method to not fail on temporary errors

shubham-pampattiwar:


internal/backup/volumesnapshotcontent_action.go
```diff
@@ -153,10 +153,11 @@ func (p *VolumeSnapshotContentBackupItemAction) Progress(operationID string, bac
 	if boolptr.IsSetToTrue(vsc.Status.ReadyToUse) {
 		progress.Completed = true
 	} else if vsc.Status.Error != nil {
-		progress.Completed = true
+		errorMessage := ""
 		if vsc.Status.Error.Message != nil {
-			progress.Err = *vsc.Status.Error.Message
+			errorMessage = *vsc.Status.Error.Message
 		}
+		p.Log.Warnf("VolumeSnapshotContent has a temporary error %s. SnapshotContent controller will retry later.", errorMessage)
```
kaovilai, line 160:
```suggestion
		p.Log.Warnf("VolumeSnapshotContent has a temporary error: %s. SnapshotContent controller will retry later.", errorMessage)
```

internal/backup/volumesnapshotcontent_action.go
```diff
@@ -153,10 +153,11 @@ func (p *VolumeSnapshotContentBackupItemAction) Progress(operationID string, bac
 	if boolptr.IsSetToTrue(vsc.Status.ReadyToUse) {
 		progress.Completed = true
 	} else if vsc.Status.Error != nil {
-		progress.Completed = true
+		errorMessage := ""
```
blackpiglet, line 156: If we don't mark the `progress.Completed` as `true` when there is an error in the VSC and VS status, one possible side-effect is that if the error cannot be resolved by retrying, the action will not be completed until the timeout. The default async operation timeout is 4 hours.

blackpiglet: Can we add a mechanism to count down the failure limitation?
For example, we can set a max try time for the VSC on error, and put that information in the VSC annotation, then update the error tried number accordingly. Until the limitation is hit, fail the action.

anshulahuja98: I am not sure how to track for number of errors, since last error might stay same and not be updated.
Might be better to somehow enforce 10min limit and then mark it as failed.

anshulahuja98 (changes requested):

internal/backup/volumesnapshotcontent_action.go
blackpiglet: Thanks. A timer is better. Propose the name as `item-operation-error-timeout`.

shubham-pampattiwar: @blackpiglet @anshulahuja98 the new parameter `--item-operation-error-timeout` sounds good, will work on a PR for velero controller and then update this PR accordingly.

anshulahuja98: this parameter however might lead to more concerns
where each plugin might want to have a different time limit, whereas this is a global setting

anshulahuja98: It is better if somehow we give this error exit control to the plugin to be smart enough to exit by it's own logic than a global setting

blackpiglet: Jus realized I didn't think it through.
It's not possible to have a global timer works in all of the plugins.
`item-operation-error-timeout` works globally, because it's trigger in the Velero server.

If we don't modify the BIA and RIA interface, we cannot pass in the timer to the Progress method.

`item-operation-error-timeout` can only works in the backup_operations_controllers.go. Looks like that is not what we want to do.

Add a default timer in the plugin code is the simplest way, but we cannot modify the timer.
Another possible way is adding the Init method just as the ObjectStore and the VolumeSnapshotter plugins.

blackpiglet: @shubham-pampattiwar 
That should work.

reasonerjt:
I think we can use a hard-coded value for the timeout or max retry, rather than expose a parameter to user to configure.
But given it's the `Progress` func in the plugin, we need to find a good way to track the time or retry number.

PR #231, opened 2024-03-18
Only generate one Async Operation for CSI backup and some code refactor

blackpiglet:
* Only generate one Async Operation for CSI backup.
* use function passed-in context.
* replace the deprecated `apimachinery/pkg/util/wait` method and variable.
* remove unneeded function parameters.

### DIFF
--- a/internal/backup/volumesnapshotcontent_action.go
+++ b/internal/backup/volumesnapshotcontent_action.go
@@ -17,26 +17,18 @@ limitations under the License.
 package backup
 
 import (
-	"context"
 	"fmt"
-	"strings"
-	"time"
 
 	snapshotv1api "github.com/kubernetes-csi/external-snapshotter/client/v7/apis/volumesnapshot/v1"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
 	"github.com/vmware-tanzu/velero-plugin-for-csi/internal/util"
 	velerov1api "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
-	"github.com/vmware-tanzu/velero/pkg/kuberesource"
-	"github.com/vmware-tanzu/velero/pkg/label"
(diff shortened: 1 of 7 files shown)


internal/backup/volumesnapshotcontent_action.go
```diff
@@ -122,43 +101,6 @@ func (p *VolumeSnapshotContentBackupItemAction) Progress(operationID string, bac
 		return progress, biav2.InvalidOperationIDError(operationID)
 	}
 
-	// The operationId is of the form <volumesnapshotcontent-name>/<started-time>
-	operationsIDParts := strings.Split(operationID, "/")
```
reasonerjt, line 126: The chunk to check if `operationID == ""` seems can also be removed?

internal/backup/volumesnapshotcontent_action.go
```diff
@@ -91,23 +84,9 @@ func (p *VolumeSnapshotContentBackupItemAction) Execute(item runtime.Unstructure
...
-		operationID = snapCont.Name + "/" + time.Now().Format(time.RFC3339)
-		itemToUpdate = []velero.ResourceIdentifier{
-			{
-				GroupResource: kuberesource.VolumeSnapshotContents,
-				Name:          snapCont.Name,
-			},
-		}
-
-	}
-
```
reasonerjt, line 110: I think we don't need to return variables for `operationID` and `itemToUpdate` since no async operation will be triggered.

internal/backup/volumesnapshotcontent_action.go
blackpiglet: Modified.

internal/backup/volumesnapshotcontent_action.go
blackpiglet: Modified.

shubham-pampattiwar (commented):
@blackpiglet Changes look good to me. But wanted to ask why the clubbing together of VS and VSC plugin operations ? 
- Were there significant performance improvements observed at scale ?
- Or anything else like ease of ux/dev ?

blackpiglet:
> @blackpiglet Changes look good to me. But wanted to ask why the clubbing together of VS and VSC plugin operations ?
> 
> * Were there significant performance improvements observed at scale ?
> * Or anything else like ease of ux/dev ?

The main consideration is not about the performance. The second factor is the main reason. There should be only one async operation for a volume backup.

reasonerjt (approved):

shubham-pampattiwar (approved):

PR #237, opened 2025-01-22
feat: support mapping the VolumeSnapshotClass via a StorageClass annotation

cfi2017:
Our use case involves two StorageClasses, one with RWO and one with RWX support. 

They both use the same provisioner, but with different arguments, and the upstream CSI requires two separate VolumeSnapshotClasses. 

We'd like to be able to specify the VolumeSnapshotClass to be used in the StorageClass directly.

An alternative here would be specifying the VolumeSnapshotClass as an annotation to each PVC, but with hundreds of PVCs managed by various different sources this is a cleaner solution IMO.

### DIFF
--- a/internal/backup/pvc_action.go
+++ b/internal/backup/pvc_action.go
@@ -125,8 +125,8 @@ func (p *PVCBackupItemAction) Execute(item runtime.Unstructured, backup *velerov
 	if err != nil {
 		return nil, nil, "", nil, errors.Wrap(err, "error getting storage class")
 	}
-	p.Log.Debugf("Fetching volumesnapshot class for %s", storageClass.Provisioner)
-	snapshotClass, err := util.GetVolumeSnapshotClass(storageClass.Provisioner, backup, &pvc, p.Log, p.SnapshotClient.SnapshotV1())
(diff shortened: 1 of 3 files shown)


sseago:
As of Velero 1.14, the velero-plugin-for-csi code has been moved to the main velero repo. Since this looks like a (perhaps slight) change in functionality, this should probably be handled as a PR against the vmware-tanzu/velero repo on the main branch. Since this repo is only used for Velero 1.13 and older, it's unlikely that a backport will be needed, as I don't anticipate any future Velero 1.13 patch releases,since Velero 1.15 is already out and we're working on the upcoming 1.16 release.

cfi2017:
> As of Velero 1.14, the velero-plugin-for-csi code has been moved to the main velero repo. Since this looks like a (perhaps slight) change in functionality, this should probably be handled as a PR against the vmware-tanzu/velero repo on the main branch. Since this repo is only used for Velero 1.13 and older, it's unlikely that a backport will be needed, as I don't anticipate any future Velero 1.13 patch releases,since Velero 1.15 is already out and we're working on the upcoming 1.16 release.

I realised this shortly after submitting. I will close this PR in favour of this one in the upstream repo:
https://github.com/vmware-tanzu/velero/pull/8646

